PR #9, opened 2015-01-27
added score social networks sl files Signed-off-by: Ioan Vranau ioan.vranau@hp.com

ioanvranau:
added score social networks sl files Signed-off-by: Ioan Vranau ioan.vranau@hp.com

### DIFF
--- a/org/openscore/slang/socialnetworks/print.sl
+++ b/org/openscore/slang/socialnetworks/print.sl
@@ -1,0 +1,9 @@
+namespace: org.openscore.slang.twitter
+operations:
+  - print:
+      inputs:
+        - text
+      action:
+        python_script: print text
+      results:
+        - SUCCESS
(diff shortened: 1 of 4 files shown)


meirwah:
need license headers for alll files
and documentation...

org/openscore/slang/socialnetworks/print.sl
```diff
@@ -0,0 +1,9 @@
+namespace: org.openscore.slang.twitter
```
meirwah: this can be a generic print action...
not related to twiiter (mybe in base/util/print?)

org/openscore/slang/socialnetworks/send_mail.sl
```diff
@@ -0,0 +1,22 @@
+namespace: org.openscore.slang.twitter
```
meirwah: use can use generic send_email operation... from here : 
https://github.com/openscore/slang-content/blob/master/org/openscore/slang/base/mail/send_mail.sl

oritstone:
Developed during the Hackathon. 
It is a great story, but we decided not to merge it

PR #18, opened 2015-02-01
added instructions for installing the CLI and running flows

SamMarkowitz:
As per #17. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,27 @@
 #slang-content
 
-##Run SLANG content
-To run SLANG content, install the SLANG CLI. For the full documentation see the [openscore](http://www.openscore.io/#/docs) website.
+score-language (Slang)
+==============
 
-###Download and Run Pre-built CLI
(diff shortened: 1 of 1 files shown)


meirwah:
why not use what i already wrote in slang project ? : https://github.com/openscore/score-language/blob/master/README.md

You are missinghere some stuff , like notice that there is a batch file fort win and for linux it's a diff file...
it look a bit weird to ask the user to go to the website , scroll down, and find the CLI zip...
There should be a download page at least...

SamMarkowitz:
I matched it to what you had in the score-language repo. I'm a little concerned about having to remember to update so many links every time there is a new version of the CLI.

meirwah:
why not have here a link to score-lang doc?

SamMarkowitz:
Do you mean remove the whole thing and replace it with just a link?

meirwah:
@orius123 what you think is better?

orius123:
I think that the docs should have a dedicated section, people don't like navigating between links.. IMHO

meirwah:
we have confilcts here...

PR #24, opened 2015-02-03
Fix less than percentage

TudorLesanHP:
Fixed. Changed the responses to:
LESS - first input is less than second input
MORE - first input is greater/equal to second input
FAILURE - input cannot be converted to int


meirwah:
@TudorLesanHP please fix conflicts...

meirwah:
:+1:

PR #28, opened 2015-02-11
Deis use case

Bonczidai:
add private key support for docker_images_maintenance use case


org/openscore/slang/base/remote_command_execution/ssh/ssh_command.sl
```diff
@@ -21,45 +21,48 @@
...
+    name: ssh_command
+    inputs:
+      - host
+      - port:
+            required: false
+      - command
+      - pty:
+            default: "'false'"
+      - username
+      - password
```
meirwah: isn't password required false also?

org/openscore/slang/base/remote_command_execution/ssh/ssh_command.sl
```diff
@@ -21,45 +21,48 @@
...
+      java_action:
+        className: org.openscore.content.ssh.actions.SSHShellCommandAction
+        methodName: runSshShellCommand
+    outputs:
+      - returnResult
+      - STDOUT
+      - STDERR
+      - exception
+    results:
+      - SUCCESS: returnCode == '0'
```
meirwah: why did you remove  (not 'Error' in STDERR) part?

org/openscore/slang/docker/images/clear_docker_dangling_images_flow.sl
```diff
@@ -29,6 +30,8 @@ flow:
     - docker_host
     - docker_username
     - docker_password
+    - private_key_file:
```
meirwah: requiered false did not work?

org/openscore/slang/base/remote_command_execution/ssh/ssh_command.sl
Bonczidai: no - because it is declared required in the content action

org/openscore/slang/base/remote_command_execution/ssh/ssh_command.sl
Bonczidai: I thought the following: being general ssh operation the only thing that matters for success is the return code (that means we don't have exceptions), we can have anything in STDERR or STDOUT - the user could interpret it as he wants.. but that's just mu approach

Let me know if you want it back..

org/openscore/slang/docker/images/clear_docker_dangling_images_flow.sl
Bonczidai: no - when we bind the inputs for a task the input expression must return a non-null value after evaluation so we have to define somewhere an empty string - the default section is perfect for that.

However it would be nice to declare defaults only in operations but I think we should change the runtime for that to pass null values to tasks (and that may not be a good idea).

.. and default implies not required which I think means that the flow deals with missing input.
This dealing involves expressions like: `private_key_file if private_key_file is not None else ""` - with the default declaration we make a shortcut for a lot of expressions like this

org/openscore/slang/base/remote_command_execution/ssh/ssh_command.sl
meirwah: sounds ok

meirwah:
@orius123 any idea why the build fails?

oritstone:
I think you should merge from master. The .Travis.yml file was added to the master

meirwah:
merge from master ? what does it mean?

orius123:
@oritstone is right, you should merge back to this branch from master in order to have the Travis yaml file that was added to master today. Otherwise Travis doesn't know how to build this project.

Bonczidai:
merged from master..

PR #29, opened 2015-02-12
Seffy

gseffy:


org/openscore/slang/docker/cadvisor/evaluate_resource_usage.sl
```diff
@@ -33,7 +33,7 @@ operation:
   name: evaluate_resource_usage
   inputs:
     - usage_query:
```
meirwah: I think "rule" is a better input name...

meirwah:
:+1:

PR #31, opened 2015-02-19
Add section on new content validation

oritstone:


oritstone:
@orius123 @SamMarkowitz please review

oritstone:
Still need to update the link to the latest validator release once it is there
@orius123 is the doc link OK, or do we have to change it also once we revive the site?

orius123:
Don't know.

SamMarkowitz:
I think you should take out the first comma and change the "should" to "must" if we really mean it.

oritstone:
fixes #31

PR #43, opened 2015-03-10
found random mistakes

SamMarkowitz:
Here are some mistakes I came across by accident while looking at the base directory. Most are just typos in the comments. One was a typo in an input name.

@meshipeer Do you think it's worthwhile for me to go through all the comments and fix up typos, grammar issues, etc? 


meshipeer:
i think this can be a good idea

PR #45, opened 2015-03-10
added generic shell commands operation

orius123:
Signed-off-by: Eliya Sadan <eliyasadan@gmail.com>


orius123:
@meirwah please review

meirwah:
sorry @orius123 .. but i think the better root for that is under base/cmd/

meirwah:
other then that.. :+1:

Issue #71, opened 2015-03-20
Move flows from : content/org/openscore/slang/docker/linux to base/os/linux


meirwah:
current flows are Linux related and not Docker specific...

PR #72, opened 2015-03-20
Moved linux flows from docker to base

tudorlesan:
Moved validate_linux_machine_ssh_access.sl, check_linux_disk_space.sl from  content/org/openscore/slang/docker/linux to content/org/openscore/slang/base/os/linux. Also changed flows that use them


tudorlesan:
@meirwah fixes #71

PR #77, opened 2015-03-23
docs changes for rename

SamMarkowitz:


meshipeer:
need to change the CLI part - and avoid version in the link

PR #80, opened 2015-03-27
Fix Openstack flows

tudorlesan:


content/io/cloudslang/openstack/list_servers.sl
```diff
@@ -39,6 +39,8 @@ flow:
         default: "'8774'"
     - openstack_username
     - openstack_password
+    - tenant_name:
+        default: "'demo'"
```
meirwah: really , demo is the defualt?

content/io/cloudslang/openstack/openstack_health_check.sl
```diff
@@ -41,9 +41,13 @@ flow:
         default: "'5000'"
     - compute_port:
         default: "'8774'"
+    - network_ID:
+        default: "''"
     - img_ref
     - username
     - password
+    - tenant_name:
+        default: "'demo'"
```
meirwah: really , demo is the defualt?

content/io/cloudslang/openstack/list_servers.sl
tudorlesan: this was something that slipped from the first implementation. I removed the hardcoded value but put it as default since all our openstack machines have a project named demo. Should I remove it and just provide it as an input everytime?

meirwah: remove it, it is not a real default value...

content/io/cloudslang/openstack/openstack_health_check.sl
meirwah: remove also this...

meirwah:
@tudorlesan did you check it works in the end? with TryStack?

tudorlesan:
@meirwah yes I checked and all of them and they work fine. You just have to provide proxy to exit HP internal network

PR #105, opened 2015-04-15
adding ping operation test

raskin-aleksandr:


raskin-aleksandr:
Test flow works locally, but fails on Travis due to ping missing error.
Pending test suits creation.

PR #107, opened 2015-04-15
adding unzip_archive operation test

raskin-aleksandr:


test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+testUnzipArchiveUNZIPFAILURE:
+  inputs:
+    - path: test_arch2
+    - out_folder: foldertest
+  description: Tests that unzip_archive.sl operation finishes with UNZIPFAILURE due to incorrect archive file name
+  testFlowPath: io.cloudslang.base.files.test_unzip_archive
+  result: UNZIPFAILURE
+
+testUnzipArchiveUNZIPFAILURE:
```
oritstone: The name of each test  case must be unique. In this case it is the same as the previous one...
Anyway, each test case should have a name that indicates what it tests, and not just failure...

test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,31 @@
...
+  inputs:
+    - path: test_arch2
+    - out_folder: \:/
+  description: Tests that unzip_archive.sl operation finishes with UNZIPFAILURE due to incorrect archive file name
```
oritstone: Description is wrong. It is copied from the test case above

test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,31 @@
...
+  testFlowPath: io.cloudslang.base.files.test_unzip_archive
+  result: UNZIPFAILURE
```
oritstone: Missing: 
- test case with missing mandatory inputs - should cause exception

raskin-aleksandr:
Unzipping operation is pending `cludslang-0.7.16` update, to test outputs on Travis

test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,56 @@
...
+####################################################
+testUnzipArchiveSuccess:
+  inputs:
+    - path: test_arch2.zip
+    - out_folder: foldertest
+  description: Tests that unzip_archive.sl operation finishes with SUCCESS when inputs are valid
+  testFlowPath: io.cloudslang.base.files.test_unzip_archive
+  result: SUCCESS
+
+testUnzipArchiveSuccess:
```
oritstone: Each test case name must be unique and specify what are we testing in this test case. This name is exactly like the one above and not informative

test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+testUnzipArchiveUNZIPFAILURE_wron_name:
+  inputs:
+    - path: test_arch2
+    - out_folder: foldertest
+  description: Tests that unzip_archive.sl operation finishes with UNZIPFAILURE due to incorrect archive file name
+  testFlowPath: io.cloudslang.base.files.test_unzip_archive
+  result: UNZIPFAILURE
+
+testUnzipArchiveUNZIPFAILURE_wron_name:
```
oritstone: Each test case name must be unique and specify what are we testing in this test case. This name is exactly like the one above and not informative

test/io/cloudslang/base/files/test_unzip_archive.inputs.yaml
```diff
@@ -0,0 +1,56 @@
...
+  inputs:
+    - path: test_arch2
```
oritstone: If the test case is about testing that the out folder is a file, you should give the path input correct (a zip file)

PR #108, opened 2015-04-16
Coreos use case

tudorlesan:


content/io/cloudslang/coreos/cluster_docker_images_maintenance.sl
```diff
@@ -62,10 +66,11 @@ flow:
                     - docker_password: coreos_password
                     - private_key_file
                     - percentage
+                    - timeout
```
meirwah: why yo needed to add timeout??

content/io/cloudslang/docker/images/clear_docker_images_flow.sl
```diff
@@ -70,6 +78,32 @@ flow:
...
+              docker_images.get_image_parents:
+                - docker_host
+                - docker_username
+                - docker_password
+                - image_name: image
+                - private_key_file
+                - timeout
+            publish:
+                - all_parent_images: >
+                    parent_images_list + " "
```
meirwah: are you sure this works????

meirwah:
Are you sure you are not trying to fetch the parent image even that it does not have a parent image???
since base image does not have parent..

content/io/cloudslang/coreos/cluster_docker_images_maintenance.sl
tudorlesan: SSHCommandAction was giving giving return_code=-1 with no error_message and the guys from content suggested increasing the timeout. It worked afterwards

tudorlesan:
@meirwah for many usecases I tried this, it worked, but probably it has to be changed to work in all cases. This particular implementation keeps overwriting parent_images, I have to change that. As for checking parent for images without parents I'm not sure how we can anticipate this. I'm not sure I understood you correctly.

meirwah:
simple question.. if you run the flow against an image with no parent.. what happens?

tudorlesan:
@meirwah I managed to pull scratch on my docker machine and tested that step for it. It works fine, if there is no parent, the image name will be "" and subtracted from the images_safe_to_delete list which doesn't change it at all.

content/io/cloudslang/docker/images/clear_docker_images_flow.sl
```diff
@@ -70,6 +85,43 @@ flow:
         publish:
           - images_list_safe_to_delete: result_set
           - amount_of_images: len(result_set.split())
+
+    - check_lists:
```
meirwah: check_lists ?? not a good name for step.. how is this related???

content/io/cloudslang/docker/images/clear_docker_images_flow.sl
```diff
@@ -70,6 +85,43 @@ flow:
...
+        do:
+          docker_utils.check_lists:
+            - all_images_list
+            - used_images_list
+        navigate:
+          BOTH_EMPTY: SUCCESS
+          USED_EMPTY: delete_images
+          NONE_EMPTY: get_parent_image
+
+    - get_parent_image:
```
meirwah: should be : get_parent_images

content/io/cloudslang/docker/images/get_image_parent.sl
```diff
@@ -0,0 +1,80 @@
...
+          docker_images.inspect_image:
+            - host: docker_host
+            - username: docker_username
+            - password: docker_password
+            - imageName: image_name
+            - privateKeyFile: private_key_file
+            - timeout:
+                required: false
+        publish:
+          - standard_out
```
meirwah: that is a bad output name.. it should be something like "image_inspect_json" or "image_info_json"...

content/io/cloudslang/docker/images/get_image_parent.sl
```diff
@@ -0,0 +1,80 @@
...
+          - standard_err
```
meirwah: why you publish standard_err if you are not using it??

content/io/cloudslang/docker/utils/check_lists.sl
```diff
@@ -0,0 +1,51 @@
...
+flow:
+  name: check_lists
+  inputs:
+    - all_images_list
+    - used_images_list
+  workflow:
+      - check_all_images:
+          do:
+            base_strings.string_equals:
+              - first_string: all_images_list
```
meirwah: why all_images_list is written here? it should be a generic op

content/io/cloudslang/docker/utils/check_lists.sl
```diff
@@ -0,0 +1,51 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
```
meirwah: All of this op, is not generic , and does not make sense...
Please use existing string_equals in parent flow instead..

meirwah:
@Bonczidai can you validate this branch works with a real CoreOS cluster?

Bonczidai:
I tested it in several cases (with cloud-slang master):

eval

it seems to work, **but** I am getting the following sporadic error (meaning: sometimes, very often actually - gives this error, then I run again with the same cluster config and it passes):

```
Error binding output: 'amount_of_images_deleted',
	Error is: Error in running script expression or variable reference, for expression: '0 if images_list_safe_to_delete == '' else amount_of_images',
	Script exception is: javax.script.ScriptException: NameError: name 'images_list_safe_to_delete' is not defined in <script> at line number 1}
```

meirwah:
@tudorlesan any idea why is that?

tudorlesan:
@meirwah I would have to see more of the execution.log. But that usually happens when a step fails and then the flow output can't be binded anymore. I'll have to check that.

PR #121, opened 2015-04-27
Fixing failing if flow running not from content root folder

raskin-aleksandr:


orius123:
please make sure the folders are deleted after the tests ran.
it stays after tests end in my env..

PR #126, opened 2015-04-29
Adding docker get_all_imges system test

raskin-aleksandr:


meirwah:
test failo, but you also have merge confilcts, try to pull first...

test/io/cloudslang/docker/images/test_get_all_images_outputs.sl
```diff
@@ -0,0 +1,102 @@
...
+          FAILURE: FAILURE
+
+    - verify_output:
+        do:
+          strings.match_regex:
+            - regex: "'hello-world'"
+            - text: list
+        navigate:
+          MATCH: delete_downloaded_image
+          NO_MATCH: VEFIFYFAILURE
```
meirwah: please note, that if the test fails.. it will not clear it's state.. @oritstone what we decide onthis in the end/?

oritstone: We did not decide yet.
We will add a flow that will clear Docker env (containers & images) and we will run it as the first task for the Docker tests.
I'm not sure if we should also make sure we clear the env at the end in case of failure...
@orius123 @meirwah WDYT?

meirwah: lets start with only the start part.. And see from there...

test/io/cloudslang/docker/images/test_get_all_images_outputs.sl
```diff
@@ -0,0 +1,102 @@
...
+  name: test_get_all_images_outputs
+  inputs:
+    - host
+    - port:
+        required: false
+    - username
+    - password
+    - image_name
+  workflow:
+    - validate_ssh:
```
meirwah: so you will need to add here a flow that clears the env...

test/io/cloudslang/docker/images/test_get_all_images_outputs.sl
```diff
@@ -0,0 +1,113 @@
...
+            - password
+        publish:
+            - list: image_list
+        navigate:
+          SUCCESS: verify_output
+          FAILURE: FAILURE
+
+    - verify_output:
+        do:
+          strings.match_regex:
```
meirwah: match_regex? can't you use equal op? or add such if missing...?

meirwah:
:+1:

PR #130, opened 2015-05-03
Fix SSH operation

oritstone:
- Add a new flow that verifies first the SSH access to the host ,and then runs the SSH command

### DIFF
--- a/content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
+++ b/content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
@@ -39,11 +39,9 @@ operation:
       - host
       - port:
             default: "'22'"
-            required: false
       - command
       - pty:
             default: "'false'"
-            required: false
       - username
       - password:
             required: false
@@ -53,13 +51,10 @@ operation:
             required: false
       - timeout:
             default: "'90000'"
-            required: false
       - characterSet:
             default: "'UTF-8'"
-            required: false
       - closeSession:
             default: "'false'"
-            required: false
       - agentForwarding:
             required: false
     action:
--- a/content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
+++ b/content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
@@ -37,24 +37,19 @@ operation:
     name: ssh_command
     inputs:
       - host
-      - port:
-            default: "'22'"
+      - port: "'22'"
       - command
-      - pty:
-            default: "'false'"
+      - pty: "'false'"
       - username
       - password:
-            required: false
+          required: false
       - arguments:
-            required: false
+          required: false
       - privateKeyFile:
-            required: false
-      - timeout:
-            default: "'90000'"
-      - characterSet:
-            default: "'UTF-8'"
-      - closeSession:
-            default: "'false'"
+          required: false
+      - timeout: "'90000'"
+      - characterSet: "'UTF-8'"
+      - closeSession: "'false'"
       - agentForwarding:
             required: false
     action:
--- a/content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
+++ b/content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
@@ -10,14 +10,14 @@
 #
 #  Inputs:
 #    - host - hostname or IP address
-#    - port - optional - port number for running the command
+#    - port - optional - port number for running the command - Default: 22
 #    - command - command to execute
-#    - pty - optional - whether to use pty - Valid: true, false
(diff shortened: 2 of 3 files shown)


content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
```diff
@@ -37,12 +38,13 @@ operation:
     inputs:
       - host
       - port:
-            required: false
+            default: "'22'"
       - command
       - pty:
             default: "'false'"
       - username
```
orius123: I think username should be required false also..

content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
```diff
@@ -0,0 +1,108 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
```
orius123: why do we need the defaults twice?

orius123:
@meirwah can you please validate and merge?

meirwah:
@orius123 tests seem to fail..

meirwah:
[ERROR] - Test case failed: testDockerStuff - Tests that it works
		Expected result: SUCCESS
		Actual result: FAILURE
[ERROR] - Test case failed: testSSHBasicallyWorks - Tests ssh actually works
		Expected result: SUCCESS
		Actual result: FAIL_VALIDATE_SSH

content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
```diff
@@ -0,0 +1,131 @@
...
+
+    outputs:
+      - returnResult
+      - standard_out
+      - standard_err
+      - exception
+    results:
+      - SUCCESS: returnCode == '0'
+      - FAILURE
+      - FAIL_VALIDATE_SSH
```
meirwah: why add a new result??? that will cause all tasks using this to wire up many navigation's?

orius123: nobody uses this content yet. any other way to report different results?

meirwah: you indicate it by the output of standard_err instead...

oritstone:
@orius123 , @meirwah 
- now that we call twice the same operation, maybe Meir is right and we do not need a different result for the first task failure.
- In addition, maybe we should do the task in a loop - first iteration should get the empty command, and second the given command?
- Regarding the doc - maybe we should leave the default values so that everyone using the flow will know what the default values are? I expect people will use the flow and not the operation, right?

meirwah:
No need for Loop here.. it will not save/improve anything...

meirwah:
agree on default  values...

content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
```diff
@@ -0,0 +1,130 @@
...
+                  required: false
+              - agentForwarding:
+                  required: false
+          publish:
+            - returnResult
+            - standard_out
+            - standard_err
+            - exception
+          navigate:
+            SUCCESS: ssh_command
```
meirwah: what does this do : SUCCESS: ssh_command???
you need to check returnCode NO?

content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
```diff
@@ -0,0 +1,130 @@
...
+            - standard_err
+            - exception
+
+    outputs:
+      - returnResult
+      - standard_out
+      - standard_err
+      - exception
+    results:
+      - SUCCESS: returnCode == '0'
```
meirwah: how will this work? where it will get the returnCode from?

content/io/cloudslang/base/remote_command_execution/ssh/ssh_flow.sl
```diff
@@ -0,0 +1,130 @@
...
+#    - agentForwarding - optional - the sessionObject that holds the connection if the close session is false
+# Outputs:
+#    - returnResult - STDOUT of the remote machine in case of success or the cause of the error in case of exception
+#    - standard_out - STDOUT of the machine in case of successful request, null otherwise
+#    - standard_err - STDERR of the machine in case of successful request, null otherwise
+#    - exception - contains the stack trace in case of an exception
+# Results:
+#    - SUCCESS - SSH access was successful and returned with code 0
+#    - FAILURE - otherwise
+#    - FAIL_VALIDATE_SSH - SSH Validation check failed
```
meirwah: FAIL_VALIDATE_SSH ?

PR #131, opened 2015-05-03
Add new flow for getting all the Docker containers

oritstone:


orius123:
Do we really need required false when we have default?

oritstone:
Was branched from fix SSH op branch. 
Can be merged only after merging #130

content/io/cloudslang/docker/containers/get_all_containers.sl
```diff
@@ -0,0 +1,97 @@
...
+            - port:
+                required: false
+            - username
+            - password:
+                required: false
+            - privateKeyFile:
+                required: false
+            - command:
+                default: >
+                  "docker ps -a -q"
```
meirwah: make the "-a" an option

PR #134, opened 2015-05-06
Clear docker host op

meirwah:


content/io/cloudslang/docker/maintenance/clear_docker_host.sl
```diff
@@ -0,0 +1,83 @@
...
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Deletes all Docker images and containers from Docker Host.
+#
+# Inputs:
+#   - docker_host - Docker machine host
```
orius123: can;t we just call it host,username and password?

content/io/cloudslang/docker/maintenance/clear_docker_host.sl
```diff
@@ -0,0 +1,83 @@
...
+                required: false
+            - private_key_file:
+                required: false
+            - timeout:
+                required: false
+        publish:
+          - all_containers: container_list
+    - clear_all_containers:
+        do:
+          docker_containers.clear_container:
```
orius123: does it also stops running containers?

content/io/cloudslang/docker/maintenance/clear_docker_host.sl
```diff
@@ -0,0 +1,83 @@
...
+            - docker_password:
+                required: false
+            - private_key_file:
+                required: false
+            - timeout:
+                required: false
+        publish:
+          - amount_of_images_deleted
+          - amount_of_dangling_images_deleted
+          - dangling_images_list_safe_to_delete
```
orius123: why do we need to publish: dangling_images_list_safe_to_delete and images_list_safe_to_delete

content/io/cloudslang/docker/maintenance/clear_docker_host.sl
meirwah: yes - it does stop+remove

content/io/cloudslang/docker/maintenance/clear_docker_host.sl
meirwah: good question...
It makes it clearer that this is the Docker host params.
Let's say this flow also needs to auth with an app, so the app param should be username/password, how will you diff between this params and the docker host one's ?

for sure we need to align this..

meirwah:
@orius123 please merge

PR #149, opened 2015-05-13
Update run_command.sl

meirwah:
small documentation fix


content/io/cloudslang/base/cmd/run_command.sl
```diff
@@ -6,7 +6,7 @@
 #   http://www.apache.org/licenses/LICENSE-2.0
 #
 ####################################################
-# This operation run a shell command
+# This operation run a shell command locally
```
SamMarkowitz: To align it with the other docs, change it to "Runs a shell command locally." - without "This operation", "runs" with an 's' and a period at the end.

PR #154, opened 2015-05-17
Adding Get_image_name_from_id docker operation test

raskin-aleksandr:


content/io/cloudslang/docker/images/get_image_name_from_id.sl
```diff
@@ -33,20 +33,20 @@ operation:
...
         default: >
             "docker images | grep " + image_id + " | awk '{print $1 \":\" $2}'"
         overridable: false
     - arguments:
-        default: "''"
+        required: false
     - characterSet:
-        default: "'UTF-8'"
+        required: false
     - pty:
```
orius123: I suggest to make this also required false instead of the default false.
same for other ssh options

test/io/cloudslang/docker/images/get_image_name_from_id.inputs.yaml
```diff
@@ -0,0 +1,19 @@
...
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+testGetImageNameFromId:
+  inputs:
+    - host: localhost
+    - port: "49153"
+    - username: root
+    - password: screencast
+    - image_id: 91c95931e552
```
orius123: maybe we should use an image version in the flow pull step in case latest will change and than the id will change

test/io/cloudslang/docker/images/test_get_image_name_from_id.sl
```diff
@@ -0,0 +1,114 @@
...
+    - get_image_name_from_id:
+        do:
+          images.get_image_name_from_id:
+            - host
+            - port
+            - username
+            - password
+            - image_id
+        publish:
+            - image_name : image_name
```
orius123: just ` - image_name` will be fine

orius123:
@raskin-aleksandr 
I have here two minor comments please fix it and we will be able to merge it back to master.
why doesn't the op use ssh flow instead of directly calling java action?

raskin-aleksandr:
Works locally for me, need to be tested at `travis-ci`

Issue #97, opened 2015-04-08
Support "dns + port + volume" in Docker run op


meirwah:
Add this as the operation inputs

PR #156, opened 2015-05-18
changing docker operation `create container` to `run container`

raskin-aleksandr:
changing docker operation `create container` to `run container` using ssh flow


meirwah:
we need to make param "name" optional..

meirwah:
try to add a test for this..
and maybe also fix #97 (this can be in a diff PR)

test/io/cloudslang/docker/maintenance/test_clear_docker_host.sl
```diff
@@ -72,14 +72,14 @@ flow:
...
-          containers.create_container:
+          containers.run_container:
             - host
             - port:
                 required: false
             - username
             - password
-            - containerName: "'xxx'"
-            - imageID: image_name_to_run
+            - container_name: "'xxx'"
```
meirwah: take this param value to the inputs file...
And if you are using it , verify it in the test...

meirwah:
:+1:

PR #157, opened 2015-05-19
get_container_metrics_cAdvisor-test

Bonczidai:


Bonczidai:
currently the build fails (locally works) - I think because the '-p 8080:8080' option is not enabled in the travis config when we start up the orius123/dind-ssh container - we need this port because cAdvisor works with REST calls

@meirwah WDYT?

orius123:
Can you configure cadvisor to work with another port?

Bonczidai:
yes, it can use any port

orius123:
Than you can expose a different port for the container with the -p param. My guess is that 8080 is already taken..

orius123:
what is this PR status?
who is assigned?

Bonczidai:
need to find out why the flow fails on circleci - locally works

Bonczidai:
another issue related to this - ssh into localhost:

if we want to ssh into localhost ( circleci env ) from flow what username & password should be used? WDYT @orius123 ?

orius123:
Not sure it will work.. we don't have the credentials and I don't know how the ssh server is configured..

Bonczidai:
the rest call fails on circleci with the message: `returnResult= failed to get Docker container "cadvisor" with error: unable to find Docker container "cadvisor"
 statusCode= 500` however the cadvisor container seems to be up

@gseffy have you encountered similar issue when you developed the flow?

gseffy:
I only got this error when the docker container is not running.
I notice in the test details that the host is localhost, did you run the flow on the same machine as the docker?

Bonczidai:
yes, the container is on the same machine as the flow (circleci env).. first I start up a cadvisor container by running a local `docker run` command then I try to get the container metrics by sending the REST call to localhost:port which is bound to the containers inner port..

gseffy:
I have something we can try, in the curl_test change the link to GET http://' + host + ':32951/api/v1.2/docker/'+cadvisor_container_name+'
it will test the rest call that the flow uses.

Bonczidai:
Fails with the same message as the operation
```
Running command: 'curl -i -H "Accept: application/json" -H "Content-Type: application/json" -X GET http://localhost:32951/api/v1.2/docker/cadvisor'
HTTP/1.1 500 Internal Server Error

Content-Type: text/plain; charset=utf-8

Date: Wed, 27 May 2015 10:40:03 GMT

Content-Length: 97



failed to get Docker container "cadvisor" with error: unable to find Docker container "cadvisor"
```

Bonczidai:
conclusion: seems that cadvisor container cannot talk to Docker environment

@orius123 do you know about any restrictions that could affect this? Also is it possible to change the docker ExecDriver version of circleci? now it is:
```
    "Driver": "overlay",
    "ExecDriver": "native-0.2",
```

and on another machine where the flow runs:
```
    "Driver": "btrfs",
    "ExecDriver": "lxc-1.0.6",
```
could this be?

WDYT?

meirwah:
@Bonczidai  it fails on both Travis and CircleCI...

Bonczidai:
`cadvisor` test suite is only activated on CircleCI, the travis error may not necessarily be related to this PR - need to check it, error is:

```
[ERROR] BUILD FAILURE
[ERROR] ------------------------------------------------------------
[ERROR] CloudSlang build for repository: "/home/travis/build/CloudSlang/cloud-slang-content" failed due to failed tests.
[ERROR] Following 1 tests failed:
[ERROR] - Error occurred while running test: test_docker_get_container_names - Tests SUCCESS case
		Error running: 'test_get_container_names'.
		 Error binding input: 'private_key_file', 
		Error is: Error in running script expression or variable reference, for expression: 'private_key_file',
		Script exception is: javax.script.ScriptException: java.lang.OutOfMemoryError: java.lang.OutOfMemoryError: PermGen space
```

orius123:
Try to merge from master. We Handled it.

Bonczidai:
still blocked here (cadvisor cannot access docker specific info)

maybe I could try to test the flow on travis (in that case we would create a cadvisor container inside the dind env - and would need to expose a port for the rest calls in travis config)

WDYT? @orius123

meirwah:
sounds like a good idea to try on travis at this stage...

Bonczidai:
updated

test/io/cloudslang/docker/cadvisor/test_get_container_metrics_cAdvisor.sl
```diff
@@ -0,0 +1,57 @@
...
+        navigate:
+          SUCCESS: validate_success_get_container_metrics_cAdvisor
+          FAILURE: C_ADVISOR_CONTAINER_STARTUP_PROBLEM
+
+    - validate_success_get_container_metrics_cAdvisor:
+        do:
+          cadvisor.get_container_metrics_cAdvisor:
+            - host
+            - cadvisor_port
+            - container: cadvisor_container_name
```
orius123: don't you want to validate that something actually returned?

test/io/cloudslang/docker/cadvisor/test_get_container_metrics_cAdvisor.sl
```diff
@@ -0,0 +1,57 @@
...
+flow:
+  name: test_get_container_metrics_cAdvisor
+
+  inputs:
+    - host
+    - cadvisor_port
+    - cadvisor_container_name
+
+  workflow:
+
```
orius123: please add a pre step that clears the env

PR #169, opened 2015-05-26
make get images use ssh flow

orius123:


orius123:
assinging @Bonczidai . can't review my own stuff...

Bonczidai:
:+1: I see that the test is already added for this, so merging..

PR #170, opened 2015-05-26
adding get_all_containers system test

raskin-aleksandr:


test/io/cloudslang/docker/containers/test_get_all_containers.sl
```diff
@@ -0,0 +1,115 @@
...
+            - username
+            - password
+            - all_containers: true
+        publish:
+          - list: container_list
+
+    - verify_list:
+        do:
+          strings.match_regex:
+            - regex: "'^.............$'"
```
meirwah: why not use "string_equals.sl" here? it should be only "hello-world"...

raskin-aleksandr: @meirwah @orius123 
`get_all_containers` return result is container ID. Do you have any idea how to get container name by it's ID?
Maybe we need one more operation?

meirwah: there is a op for that :  get_container_names

raskin-aleksandr: `get_container_names` returns all containers names, i cannot get specific ID's name.

meirwah: but since you run here(this test) ONLY 1 container, there should be only 1 name, no?

raskin-aleksandr: Yes, i will get only 1 name.
But this does not help me to test `get_all_containers` operation

orius123:
```shell
docker ps -a | tail -n +2 | awk '{print $2}'
```
should work

orius123: just now i got your problem...
this works

```yaml
    - verify_list:
        do:
          strings.string_equals:
            - first_string: >
                len(list.rstrip().split(" "))
            - second_string: 2
```

PR #178, opened 2015-06-11
Cluster docker images maintenance test

Bonczidai:
this test needs to be run on the **local** machine (because we don't have a CoreOS cluster env in the form of Docker images)

what it does:
    - retrieves the ip adresses of the machines in the cluster
    - cleanup on the machines (so they will not contain any images)
    - prepares a used and an unused Docker image
    - runs the flow
    - verifies only one image remained in the cluster
    - delete the used image

### DIFF
--- a/content/io/cloudslang/docker/images/clear_docker_images.sl
+++ b/content/io/cloudslang/docker/images/clear_docker_images.sl
@@ -61,7 +61,8 @@ flow:
         do:
           ssh.ssh_flow:
             - host
-            - port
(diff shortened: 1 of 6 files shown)


orius123:
please add how and where to run the flow to the flow header.
or create a testcase with the instructions and a test suite that we don't run (i.e. `coreos`)

Bonczidai:
added test suite + description

Bonczidai:
tested locally

PR #181, opened 2015-06-13
Add docker build image operation + test

### LINKED ISSUE #96: Add Docker build op
should be named build-image op.
reside in the images folder...

Bonczidai:
resolves #96 

### DIFF
--- a/content/io/cloudslang/docker/images/build_image.sl
+++ b/content/io/cloudslang/docker/images/build_image.sl
@@ -9,20 +9,19 @@
 # Builds a Docker image based on a Dockerfile.
 #
 # Inputs:
-#   - docker_user - optional - Docker username of the created image from a construct like 'docker_user/image_name:tag' - Default: no username
-#   - image_name - name of the created image from a construct like 'docker_user/image_name:tag'
-#   - tag - optional - tag of the created image from a construct like 'docker_user/image_name:tag' - Default: latest
-#   - workdir - optional - path to the directory that contains the Dockerfile
+#   - docker_image - Docker image specifier e.g. 'docker_user/image_name:tag'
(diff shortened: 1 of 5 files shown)


content/io/cloudslang/docker/images/build_image.sl
```diff
@@ -0,0 +1,87 @@
...
+####################################################
+
+namespace: io.cloudslang.docker.images
+
+operation:
+  name: build_image
+  inputs:
+    - docker_user:
+        default: "''"
+    - image_name
```
orius123: You should use required false.

orius123:
You should use required false instead of specifying defaults. 
And please use the new ssh flow instead of the ssh action directly.

content/io/cloudslang/docker/images/build_image.sl
```diff
@@ -0,0 +1,122 @@
...
+    - password:
+        required: false
+    - private_key_file:
+        required: false
+    - character_set:
+        required: false
+    - pty:
+        required: false
+    - timeout:
+        default: "'3000000'"
```
Bonczidai: added here default for timeout because docker build operation needs higher timeout value than the default for the ssh flow

test/io/cloudslang/docker/images/test_build_image.sl
```diff
@@ -0,0 +1,183 @@
...
+flow:
+  name: test_build_image
+  inputs:
+    - docker_image
+    - base_image:
+        default: "'busybox:latest'"
+    - workdir:
+        default: "'.'"
+    - dockerfile_name:
+        default: "'Dockerfile'"
```
Bonczidai: we need here defaults because other tasks in the flow (not only the build image flow) use these values

PR #186, opened 2015-06-16
Added create_job operation (Jenkins)

tudorlesan:


content/io/cloudslang/jenkins/create_job.sl
```diff
@@ -0,0 +1,86 @@
...
+          <blockBuildWhenDownstreamBuilding>false</blockBuildWhenDownstreamBuilding>
+          <blockBuildWhenUpstreamBuilding>false</blockBuildWhenUpstreamBuilding>
+          <triggers/>
+          <concurrentBuild>false</concurrentBuild>
+          <builders/>
+          <publishers/>
+          <buildWrappers/>
+        </project>'''
+
+        job = jobs.create(job_name, CONFIG_XML)
```
orius123: Why not to use `jenkins.EMPTY_CONFIG_XML` like in their examples?
https://python-jenkins.readthedocs.org/en/latest/example.html

tudorlesan: I tried that. The library we have does not have any "EMPTY_CONFIG_XML". So we would have to find another version of it. The second reason is that I need to create a job that contains a scm, otherwise the flow fails, this xml solves that. I know it looks weird in the operation, could we extract this xml someplace else?

orius123: Try to make the library work. If it doesn't I think we should load the xml from an actual file.

tudorlesan: Even if I make the library work. The EMPTY_CONFIG_XML will lead in flow failure since it will create a job without a scm. A solution to this would be to add another step which adds a scm to a job but I looked over the functions and they don't provide an "addSCM" just "updateSCM" which works on jobs that are created from the beginning with SCM.

tudorlesan: The point is, after reading a bit about it, a much better way would be to use the REST API but that would mean changing the whole flows. This library has a lot of functions that don't work properly or at all. I think it's not developed anymore.

orius123: If you need it for the tests, go ahead and do it with "home-made" http calls. But I don't think we want to extend this integration any more for now.

orius123: Another way around this is to request the xml path as an input and provide one for the test.

tudorlesan: Sounds good

content/io/cloudslang/jenkins/create_job.sl
```diff
@@ -0,0 +1,54 @@
...
+#
+####################################################
+# Creates a Jenkins job
+#
+# Prerequisites: jenkinsapi Python module
+#
+# Inputs:
+#   - url - URL to Jenkins
+#   - job_name - name of job to create
+#   - config_xml - configuration xml used to create a jenkins job, actual file must be passed not its path
```
orius123: when you say actual file you mean a string?

content/io/cloudslang/jenkins/create_job.sl
```diff
@@ -0,0 +1,54 @@
...
+####################################################
+
+namespace: io.cloudslang.jenkins
+
+operation:
+  name: create_job
+  inputs:
+    - url
+    - job_name
+    - config_xml
```
tudorlesan: Extracted the config_xml as input and tested it. It expects the whole file though, not its path. That means that this operation must only be used within a flow in conjunction with read_from_file.sl which will pass its output to config_xml.

content/io/cloudslang/jenkins/create_job.sl
tudorlesan: yes

orius123: I thing that a path or a string is better.
this way we can pass it as an input

PR #190, opened 2015-06-17
Use ssh_flow in get_dangling_images & fix docs

oritstone:


oritstone:
@orius123 you can merge if it passes the build

PR #199, opened 2015-06-21
Remove the expression from flow's result since they are not supported for flows

oritstone:


content/io/cloudslang/docker/images/get_all_images.sl
```diff
@@ -87,8 +88,36 @@ flow:
             - returnResult
             - return_code
             - standard_err
+
+    - verify_no_error_in_stderr:
```
orius123: ssh command already checks for this

PR #205, opened 2015-06-23
consul decode key from base 64

gseffy:


meirwah:
@gseffy  can you please update this...

tudorlesan:
@meirwah her change was needed because, when getting a consul key, it would be encrypted. But I already used this change in the consul tests branch, so I think this can be deleted.

meirwah:
ok...

Issue #180, opened 2015-06-11
check_linux_disk_space operation actual behavior different than expected ?


Bonczidai:
`check_linux_disk_space` op is supposed to "Check the disk space percentage on a Linux machine"

current command:  df -kh | grep -v "Filesystem" | awk 'NR==1{print $5}'
current returned value: first line (red rectangle on the screenshot)
current behavior: disk space of the first listed filesystem

I would go with the following change:

command:  df -kh . | grep -v "Filesystem" | awk '{print $5}'
returned value: 5% (green rectangle)
behavior: disk space of the partition the current directory belongs to (the entry point for ssh)

![df](https://cloud.githubusercontent.com/assets/8058961/8109887/ebc5f8fc-1061-11e5-8641-3bc8df90c1fd.png)

WDYT?

orius123:
Y don't we get the mount / partition as an input?
or take `root``/` if none given

Bonczidai:
:+1:

tudorlesan:
Solved by #207

PR #207, opened 2015-06-23
fix check linux disk space

orius123:


Bonczidai:
this also fixes #180 , right? ..

orius123:
actually it sort of does (didn't think about it..) but now that you say it, I will add an input for the mount to check.

Bonczidai:
:+1: 

added the following changes:
- moved mount input from ssh flow to check_linux_disk_space flow
- removed some default values in the description section
- trimmed the percentage output so it will contain `x%` without newline

also tested the flow on an ubuntu machine.. please check it and merge

Issue #183, opened 2015-06-15
Content system test coverage - Local tests


Bonczidai:
The following tests need to be run (for now) on the local machine:

- [x] cAdvisor @Bonczidai (cadvisor container cannot talk to Docker env) - test suite: cadvisor_local
  - [x] evaluate_resource_usage
  - [x] get_container_metrics_cAdvisor
  - [x] parse_cadvisor_container
  - [x] report_container_metrics_cAdvisor
  - [x] restart_container_base_on_usage
- [x] coreos @Bonczidai (env issue) - test suite: coreos_local
  - [x] cluster_docker_images_maintenance
  - [x] get_machine_public_ip
  - [x] list_machines_id
  - [x] list_machines_public_ip
- [x] swarm @Bonczidai (env issue) - test suite: swarm_local
  - [x] clear_swarm_cluster
  - [x] get_cluster_info
  - [x] run_container_in_cluster
  - [x] demo_containers
  - [x] test_add_node_to_cluster
- [x] report_mysql_status @orius123 test suite: mysql-local
- [x] demo-dev-ops @Bonczidai (disk space issue) - test suite: docker_local
- [x] marathon @oritstone test suite: marathon_local
- [x] ping_hosts @oritstone test suite: mail-local
- [x] linux OS @oritstone test suite: linux-local
  - [x] restart_process
  - [x] restart_server
  - [x] restart_service
- [x] openstack @oritstone test suite: openstack-local
- [x] http_client_action @oritstone - no test
- [x] demo_create_app_and_send_mail @oritstone - no test

PR #215, opened 2015-06-29
add test for cadvisor:restart_container_base_on_usage

Bonczidai:

### DIFF
--- a/test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
+++ b/test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
@@ -8,7 +8,10 @@
 ########################################################################################################
 test_restart_container_base_on_usage_success:
   testSuites: [cadvisor_local]
-  description: Tests SUCCESS case - this test needs to be run on the local machine by activating the test suite and supplying the inputs
+  description: >
+    Tests SUCCESS case - starts a cadvisor container, waits for the container to start up and runs
+    the restart_container_base_on_usage flow against it. The Docker host is cleared before and after
+    the actual test logic.
   testFlowPath: io.cloudslang.docker.cadvisor.test_restart_container_base_on_usage
   inputs:
     - cadvisor_port: "23569"
--- a/test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
+++ b/test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
@@ -1,0 +1,19 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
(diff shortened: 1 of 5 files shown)


test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
```diff
@@ -0,0 +1,19 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+########################################################################################################
+test_restart_container_base_on_usage_success:
+  testSuites: [cadvisor_local]
```
meirwah: why is this local?

test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
```diff
@@ -0,0 +1,19 @@
...
+  description: Tests SUCCESS case - this test needs to be run on the local machine by activating the test suite and supplying the inputs
```
meirwah: this is not a description, please desc what you test here? what logic you look you want to see here...

test/io/cloudslang/docker/cadvisor/test_restart_container_base_on_usage.sl
```diff
@@ -0,0 +1,132 @@
...
+            - cadvisor_port
+            - machine_connect_port:
+                default: port
+                required: false
+            - username
+            - password:
+                required: false
+            - privateKeyFile:
+                default: private_key_file
+                required: false
```
meirwah: no navigate here?
where is the test logic here?

test/io/cloudslang/docker/cadvisor/restart_container_base_on_usage.inputs.yaml
Bonczidai: we had issues with cadvisor container (couldn't talk to the docker env) so the get_container_metrics use case did not work on circleci (so these tests are noted in #183 and we'll be resolved in a later stage)

.. however we may try it one travis to see if it has the same problem

test/io/cloudslang/docker/cadvisor/test_restart_container_base_on_usage.sl
Bonczidai: I followed here the convention: use default navigation for the task that calls the flow that should be tested, all the other tasks produce specific results on failure - this would mean that the flow ends with FAILURE only if the tested flow ended with FAILURE (not other tasks like clearing the host)

do you think it is better to declare specific result for this task? like `CALL_RESTART_CONTAINER_BASE_ON_USAGE_PROBLEM`

meirwah: I think so... so we can tell on CI which step failed easily...

Bonczidai:
updated - circleci is still having problems

PR #216, opened 2015-06-30
Aligned diskspace_healthc_check unrequired inputs

tudorlesan:

### DIFF
--- a/content/io/cloudslang/base/os/linux/diskspace_health_check.sl
+++ b/content/io/cloudslang/base/os/linux/diskspace_health_check.sl
@@ -32,9 +32,11 @@ flow:
   inputs:
     - docker_host
     - docker_username
-    - docker_password
-    - private_key_file:
+    - docker_password:
         default: "''"
+        required: false
+    - private_key_file:
+        required: false
     - percentage
     - timeout:
         required: false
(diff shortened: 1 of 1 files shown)


content/io/cloudslang/base/os/linux/diskspace_health_check.sl
```diff
@@ -32,9 +32,11 @@ flow:
   inputs:
     - docker_host
     - docker_username
-    - docker_password
-    - private_key_file:
+    - docker_password:
         default: "''"
```
meirwah: @oritstone do we need default: "''"? if we do  required: false ?

oritstone: We do not need the empty default value. It should be removed. 
And then in the tasks, the input should change to required: false

content/io/cloudslang/base/os/linux/diskspace_health_check.sl
```diff
@@ -32,9 +32,11 @@ flow:
   inputs:
```
oritstone: In the documentation above, the password & private key file should be documented as optional

oritstone:
Closing. Working on it in #223

PR #218, opened 2015-06-30
cadvisor: fix sporadic failure by adding sleep

Bonczidai:


meirwah:
:+1:

PR #223, opened 2015-07-01
Remove empty default value from docker images & maintenance flows

oritstone:


oritstone:
@orius123, @meirwah can you have a look? The tests pass locally, but fail on Travis

PR #229, opened 2015-07-05
Bugfix flow restart service

lirantal:
1. Handling of errors/success isn't correctly detecting the actual status of the service restart command. I altered it to accommodate the real status when working in Linux environments.
2. Updated documentation for the flow

### DIFF
--- a/content/io/cloudslang/base/os/linux/restart_service.sl
+++ b/content/io/cloudslang/base/os/linux/restart_service.sl
@@ -52,7 +52,7 @@ flow:
             - port:
                 required: false
             - sudo_command: "'echo -e ' + password + ' | sudo -S ' if bool(sudo_user) else ''"
-            - command: "sudo_command + 'service ' + service_name + ' restart'"
+            - command: "sudo_command + 'service ' + service_name + ' restart' + ' && echo CMD_SUCCESS'"
             - username
             - password:
                 required: False
@@ -68,7 +68,7 @@ flow:
         do:
           strings.string_occurrence_counter:
             - string_in_which_to_search: standard_out
-            - string_to_find: "'done'"
+            - string_to_find: "'CMD_SUCCESS'"
         navigate:
           SUCCESS: SUCCESS
           FAILURE: FAILURE
--- a/content/io/cloudslang/base/os/linux/restart_service.sl
+++ b/content/io/cloudslang/base/os/linux/restart_service.sl
@@ -13,6 +13,7 @@
 #       - username - username to connect as
 #       - password - password of user
 #       - service_name - linux service name to be restarted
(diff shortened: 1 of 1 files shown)


meirwah:
On which Linux distro did you test it?
when I run this flow (with your change) the flow ends with failure, with the next msg:
standard_err = stdin: is not a tty

lirantal:
@meirwah it's on my ubuntu 14.04 
also, if you take a look at the change, I only referred to ```standard_out``` instead of ```standard_err``` which ```standard_out``` is actually available to me through ```publish``` section.

I'm wondering - can you try running the exact same flow on your machine with the 'master' branch and see if you're getting a different response? also which linux dist and how are you running the flow?

lirantal:
@meirwah hey, thanks for looking into this with me today, I found the issue that was confusing us - I used "sudo=1" in the parameters instead of "sudo_user=1" and that's why it was always failing.

Anyway, here you can see a (partial) screenshot of the flow working correctly:

image

can you provide any information on why this doesn't work on your Linux?

meirwah:
@Bonczidai  can you check this flow with another Linux distro?

lirantal:
I also have another workaround to suggest which is to modify the command to
explicitly echo output on success.

For example :
```
service apache2 restart & & echo SUCCESS
```
then we can inspect stdout for that string.

What do you think?

meirwah:
makes sense!

lirantal:
Great.

I'll update the PR.

orius123:
What about using '<command> && exit 0'  and than checking the exit code?

lirantal:
Sadly there's an issue with the return code parameter. If you can fix it
that's great, otherwise we need some safe workaround there.

orius123:
What issue?

lirantal:
@meirwah I updated the flow, check it out.

@orius123 I could be working with the return_code parameter incorrectly, but take a look at this example with the restart_service flow:

I changed the ```restart_service``` flow to use the return_code that the SSH command from the imported script exposes:
image

Then I ran the flow with arguments that I know will make it fail:
image

As you can see return_code is 0 even though the flow failed.
Am I doing something wrong with the return_code usage in the flow or is that an actual defect?

orius123:
I would expect the return code to return code of the command, so I think it's a defect. @meirwah can you look into it?

lirantal:
@orius123 indeed, that's why I was suggesting a safe-bet workaround at least until this issue is solved :-)

orius123:
:+1:

Bonczidai:
checked the flow on Fedora 22 - works :+1:

lirantal:
cool, thanks for confirming! :)

Issue #236, opened 2015-07-06
Solution to flow result evaluation problem - when using ssh flow


Bonczidai:
Should find a solution to result evaluation problem in case of recently promoted operations that are now flows and use the SSH flow - usually we need to check if some keywords are found in the return result (or stdout stderr etc.)

Now we have to add multiple tasks to check the occurrences of such strings - need a more elegant way

maybe add a checking feature to the SSH flow or add an operation that can search for multiple keywords in a string (this way we need only one task to check the result)

other ideas, opinions?

orius123:
I agree that this an issue. but don't you think that we should solve it with a language feature instead of content one?

Bonczidai:
:+1:  may be a good idea - we will need to do a brainstorming to hear everybody's opinions on this

oritstone:
I think that #420 is improving the situation

Bonczidai:
CloudSlang/cloud-slang#569 will solve this issue

tethryus:
This is a feature that will be handled at language level.

PR #237, opened 2015-07-06
update CoreOS use case (use ssh flow, remove default values)

### LINKED ISSUE #225: Remove empty default values from CoreOs flows
Remove empty values from required inputs in:

get_machine_public_ip
list_machines_id
list_machines_public_ip

Bonczidai:
Fixes #225 

### DIFF
--- a/content/io/cloudslang/coreos/get_machine_public_ip.sl
+++ b/content/io/cloudslang/coreos/get_machine_public_ip.sl
@@ -11,14 +11,16 @@
 # Inputs:
 #   - machine_id - ID of the machine
 #   - host - CoreOS machine host; can be any machine from the cluster
-#   - port - optional - SSH port - Default: 22
+#   - port - optional - SSH port
 #   - username - CoreOS machine username
-#   - password - CoreOS machine password; can be empty since CoreOS machines use private key file authentication
-#   - privateKeyFile - optional - path to the private key file - Default: none
-#   - arguments - optional - arguments to pass to the command - Default: none
-#   - pty - whether to use PTY - Valid: true, false - Default: false
-#   - timeout - time in milliseconds to wait for the command to complete - Default: 90000
-#   - closeSession - if false SSH session will be cached for future calls of this operation during life of the flow, if true SSH session used by this operation will be closed - Valid: true, false - Default: false
+#   - password - optional - CoreOS machine password; can be empty since CoreOS machines use private key file authentication
+#   - private_key_file - optional - path to the private key file
+#   - arguments - optional - arguments to pass to the command
+#   - character_set - optional - character encoding used for input stream encoding from target machine - Valid: SJIS, EUC-JP, UTF-8
+#   - pty - whether to use PTY - Valid: true, false
+#   - timeout - time in milliseconds to wait for the command to complete
+#   - close_session - if false SSH session will be cached for future calls of this operation during life of the flow, if true SSH session used by this operation will be closed - Valid: true, false
+#   - agent_forwarding - optional - whether to forward the user authentication agent
 # Outputs:
 #   - public_ip: public IP address of the machine based on its ID
 # Results:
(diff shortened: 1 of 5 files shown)


Bonczidai:
fixes #225

Bonczidai:
circleci build is not running:

```
Minor Service Outage - components affected: CircleCI (partial_outage), GitHub (operational), Heroku (operational), AWS (operational), Pusher Pusher REST API (operational), Pusher WebSocket client API (operational)

major incident created at 2015-07-06T15:35:10.629-07:00 - updated at 2015-07-06T21:50:34.944-07:00
Builds dequeueing are going very slowly because of DB performance issue. We are now investigating what causes the performance issue.
```

meirwah:
It seems CircleCI is up again right?

Bonczidai:
yes - looked into the logs (seems OK)

oritstone:
@Bonczidai since the test suites runs only locally, did you verify the content is still functional after the changes?

content/io/cloudslang/coreos/get_machine_public_ip.sl
```diff
@@ -11,14 +11,15 @@
...
-#   - privateKeyFile - optional - path to the private key file - Default: none
-#   - arguments - optional - arguments to pass to the command - Default: none
-#   - pty - whether to use PTY - Valid: true, false - Default: false
-#   - timeout - time in milliseconds to wait for the command to complete - Default: 90000
-#   - closeSession - if false SSH session will be cached for future calls of this operation during life of the flow, if true SSH session used by this operation will be closed - Valid: true, false - Default: false
+#   - password - optional - CoreOS machine password; can be empty since CoreOS machines use private key file authentication
+#   - private_key_file - optional - path to the private key file
+#   - arguments - optional - arguments to pass to the command
+#   - character_set - optional - character encoding used for input stream encoding from target machine - Valid: SJIS, EUC-JP, UTF-8
+#   - pty - whether to use PTY - Valid: true, false
```
oritstone: pty, timeout, close_session - optional
agent_forwarding - missing

Bonczidai:
of course - verified :wink:

content/io/cloudslang/coreos/get_machine_public_ip.sl
```diff
@@ -28,41 +29,88 @@
...
+        required: false
     - timeout:
-        default: "'90000'"
-    - closeSession:
-        default: "'false'"
-    - agentForwarding:
+        required: false
+    - close_session:
+        required: false
+    - agent_forwarding:
```
oritstone: I think that agent_forwarding be required: false, without default value & overridable true

content/io/cloudslang/coreos/get_machine_public_ip.sl
```diff
@@ -28,41 +29,88 @@
...
+            - string_to_find: "'ssh-agent'"
+        navigate:
+          SUCCESS: FAILURE
+          FAILURE: check_unable_in_stderr
+
+    - check_unable_in_stderr:
+        do:
+          strings.string_occurrence_counter:
+            - string_in_which_to_search: standard_err
+            - string_to_find: "'Unable'"
```
oritstone: In the original content, the verification was for the string "ERROR"
Was it a mistake? You changes it to "Unable"

content/io/cloudslang/coreos/list_machines_id.sl
```diff
@@ -9,16 +9,17 @@
...
-#   - arguments - optional - arguments to pass to the command - Default: none
-#   - characterSet - optional - character encoding used for input stream encoding from target machine - Valid: SJIS, EUC-JP, UTF-8 - Default: UTF-8
-#   - pty - optional - whether to use pty - Valid: true, false - Default: false
-#   - timeout - optional - time in milliseconds to wait for command to complete - Default: 90000
-#   - closeSession - optional - if false SSH session will be cached for future calls of this operation during the life of the flow, if true SSH session used by this operation will be closed - Valid values: true, false - Default: false
+#   - password - optional - CoreOS machine password; can be empty since CoreOS machines use private key file authentication
+#   - private_key_file - optional - path to the private key file
+#   - arguments - optional - arguments to pass to the command
+#   - character_set - optional - character encoding used for input stream encoding from target machine - Valid: SJIS, EUC-JP, UTF-8
+#   - pty - whether to use PTY - Valid: true, false
```
oritstone: pty, timeout, close_session -> optional

content/io/cloudslang/coreos/list_machines_id.sl
```diff
@@ -28,38 +29,81 @@
...
+          - machines_id_list: >
+              returnResult.replace("\n"," ").replace("MACHINE ","",1).replace("...", "")[:-1]
+              if (return_code == '0' and (not 'ERROR' in standard_err)) else ''
+          - standard_err
+
+    - check_error_in_stderr:
+        do:
+          strings.string_occurrence_counter:
+            - string_in_which_to_search: standard_err
+            - string_to_find: "'ERROR'"
```
oritstone: This is already verified by the ssh_command inside the ssh_flow, so I think it is redundant

content/io/cloudslang/coreos/get_machine_public_ip.sl
Bonczidai: I would say that in this case the agent forwarding is mandatory since the command is *fleetctl ssh* (ssh in ssh) - the machine which was accessed may execute a command on another machine in the cluster

without agent-forwarding you (probably) get: 
```
Failed building SSH client: SSH_AUTH_SOCK environment variable is not set. Verify ssh-agent is running. See https://github.com/coreos/fleet/blob/master/Documentation/using-the-client.md for help.
```

content/io/cloudslang/coreos/get_machine_public_ip.sl
Bonczidai: yes, it covers the case when an invalid machine ID is given (found out when testing the flow)

content/io/cloudslang/coreos/get_machine_public_ip.sl
oritstone: So do we still need the close_session param?

content/io/cloudslang/coreos/list_machines_id.sl
Bonczidai: in the ssh command the 'Error' keyword is verified (case sensitive because of python `in` construct) but the verficiation need to be case insensitive in this case (`string_occurrence_counter` does this) because the string in the error message will be 'ERROR' with uppercase

```
2014/09/18 17:10:50 INFO client.go:278: Failed getting response from http://127.0.0.1:4001/: dial tcp 127.0.0.1:4001: connection refused
2014/09/18 17:10:50 ERROR client.go:200: Unable to get result for {Get /_coreos.com/fleet/machines}, retrying in 100ms
2014/09/18 17:10:50 INFO client.go:278: Failed getting response from http://127.0.0.1:4001/: dial tcp 127.0.0.1:4001: connection refused
2014/09/18 17:10:50 ERROR client.go:200: Unable to get result for {Get /_coreos.com/fleet/machines}, retrying in 200ms
```

content/io/cloudslang/coreos/get_machine_public_ip.sl
Bonczidai: why would we not need it? :eyes: multiple *fleetctl ssh* commands can be executed in the same session

content/io/cloudslang/coreos/list_machines_id.sl
oritstone: So maybe the correct way is to do the verification in the ssh_flow...

content/io/cloudslang/coreos/get_machine_public_ip.sl
oritstone: :+1:

content/io/cloudslang/coreos/get_machine_public_ip.sl
oritstone: :+1:

content/io/cloudslang/coreos/list_machines_id.sl
Bonczidai: ok,

however I don't like putting any condition in the ssh command - IMHO the condition should be only `returnCode == '0'` and let the parent flows decide what FAILURE means - what if I want an ssh call that is expected to cause such a situation that 'Error' is expected in the stderr

this may also be related to #236

oritstone: I think you are right, we can address it later on when we deal with #236 
For now a lot of our content expects this check inside the ssh_command... 
So leave it like this so far until we deal with it

Bonczidai:
pushed my updates ..

PR #242, opened 2015-07-07
updating README with explanation text about command line arguments

lirantal:
Makes it easier for new-comers to adopt the tool and understand command line arguments when working with cslang console for the first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,14 @@ CloudSlang is a YAML based language for writing human-readable workflows for the
 4. Run the executable :
   - For Windows : cslang.bat 
   - For Linux : bash cslang
-5. Run a simple example print text flow:  run --f ../content/io/cloudslang/base/print/print_text.sl --i text=first_flow
+5. Run a simple example print text flow: run --f ../content/io/cloudslang/base/print/print_text.sl --i text=first_flow --cp ../content/
+
+```
+Command line arguments in the example:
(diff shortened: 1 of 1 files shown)


meirwah:
Hey,
we do have "help" capability in the CLI, so users can see the CMD arguments... do you think it's not enough?

I usually do not see other OSS project writing CLI arguments in the README (https://github.com/docker/docker, https://github.com/StackStorm/st2) but they usually put it in the website docs (which we have under : http://www.cloudslang.io/#/docs#cloudslang-cli )

WDYT?

lirantal:
The intention was not to cover all available parameters, but rather to update the final step of running a flow and explaining in a nutshell about the flags so they make sense.

It took me a little while until I figured out the need for the classpath, and it wasn't all that obvious.
I also didn't download the archive but rather just cloned the respective repositories which is what IMHO a lot of people will do.

meirwah:
but in this case.. adding the "--cp ../content/" can be tricky (since it depends where did you download the content repo to).

maybe we need to add a different section for building from source...

lirantal:
You're right, that could be confusing.
Would you like me to work out another section of the README that will go through the steps of building from source for the 2 repositories? (the CLI and the content)

meirwah:
:+1: 
That will be great!

PR #246, opened 2015-07-12
Unignore clear containers wrapper test

### LINKED ISSUE #235: Test ignored: testClear_containers_wrapper_Success

oritstone:
Fixes #235 

### DIFF
--- a/content/io/cloudslang/docker/containers/examples/demo_clear_containers_wrapper.sl
+++ b/content/io/cloudslang/docker/containers/examples/demo_clear_containers_wrapper.sl
@@ -13,7 +13,8 @@
 #   - linked_container_ID - ID of the linked container
 #   - docker_host - Docker machine host
 #   - docker_username - Docker machine username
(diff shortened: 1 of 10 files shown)


Bonczidai:
disk space issue again (or still) - application image is too big, so I did the following workaround:
now the test flow uses `busybox` for the linked container, other logic remains the same. linked image is a flow input for the test with default value `meirwa/spring-boot-tomcat-mysql-app` so the only thing we have to do in the future (when we will have enough disk space) is to remove the explicit inputs from the test suite.

WDYT?

meirwah:
from one hand it's a bit weird...
since the flow purpose is the clear the demo env(which uses meirwa/spring-boot-tomcat-mysql-app image not busybox) , so what are we testing here?

On the other hand, let's take into account this is an example flow, and further more , it's a cleanup to a specific example flow, so I would invest more on this test...

meirwah:
bottom line.. i'm merging this...

PR #248, opened 2015-07-13
align_marathon_names [1]

### LINKED ISSUE #58: marathon folder - flow output names
The style of names didn't match in the comments and code. I believe the code style was correct of all the operations because they came from a Java action. (I changed the comments accordingly.) But I think the style in the flows should be changed to match the style that was found in the comments.

### LINKED ISSUE #347: Marathon enhancements
Currently the marathon env is set up using `test/io/cloudslang/marathon/setup_marathon.sl` accorind to http://viralkitty.com/marathon-and-docker/ - based on Docker containers on a regular Docker machine.

Marathon has two types of containerizers: default and Docker. The former setup only works for default containerizer (no Docker apps) - theoretically should work with Docker also but we had problems when developing the flow.

The following setup ( http://www.livewyer.com/blog/2015/02/16/running-apache-mesos-inside-docker-coreos ) is similar - but for CoreOS machine -- this one works with Docker containerizer without problems.

We should compose a setup flow for CoreOS machine, and add  a parent flow where you could select between setup type: regular / CoreOS. For the build we should use the CoreOS one (we run the flows on CoreOS machine anyway).

The CoreOS setup seem to work much better than the regular one (this may solve build problems regarding marathon -- which are hard to debug: sporadic, need to ssh into the build etc.), we could also use Docker sample app for marathon flows - `test/io/cloudslang/marathon/sample-marathon-app.json` (e.g. we could use `content/io/cloudslang/base/network/verify_app_is_up.sl`) to test whether the flow is up (much professional than a `sleep` -- in `test/io/cloudslang/marathon/test_create_app.sl`)

we should also extract the setup flows in content folder - not tests ( #289  )

long issue (fixes will be shorter), but these are the necessary details.. :baby_chick:

### LINKED ISSUE #376: Unignore Marathon Tests
failing at a high rate, ignored temporary 
  - test/io/cloudslang/marathon/create_app_inputs.yaml
  - test/io/cloudslang/marathon/update_app_inputs.yaml

SamMarkowitz:
Fixes #58, Fixes #347, Fixes #376


orius123:
Didn't we decide going lower_case all over?
I would get rid from the proxy configuration and let the user add it if he needs. @oritsyone WDYT?

Bonczidai:
I would also go with snake_case all over - we should establish a convention - so let's hear what others think .. @oritstone @meirwah @SamMarkowitz @tudorlesan 

I think that in all the cases (except input for Java action) is more natural to use lower case

in case of Java action inputs we could (should we?) use something like this:
```
- input_name
- inputName:
    default: input_name
    overridable: false
```

tudorlesan:
a similar thing can be done for outputs like returnResult, statusCode etc. Just have the outputs like:
```
outputs:
 - return_result: returnResult
 - status_code: statusCode
```
etc.

orius123:
@Bonczidai WDYT of:
`InputName: input_name` :question: 
and than you can pass both.

Bonczidai:
@orius123 :+1: I kinda like that

Bonczidai:
so, are we going to adapt the all lowercase convention - are there objections?

oritstone:
I agree. We should go with lower case

Bonczidai:
:+1:  - @SamMarkowitz please update this PR to use all lowercase inputs / outputs

SamMarkowitz:
Just want to make sure I understand before I go ahead. 

In the operation that uses the Java action I will use `inputName: input_name`. Then in all flows that use that operation I will change `inputName` to `input_name`.

Is that correct @orius123, @oritstone, or @Bonczidai ?

oritstone:
:+1:

SamMarkowitz:
Two questions:
1
------
I don't think it's going to work that way because of inputs that have default values already or are not required.

For example, in the operation `delete_app` one of the current inputs looks like this:

```yaml
- proxyHost:
        required: false
``` 

I can't do something like this:

```yaml
- proxyHost: proxy_host
        required: false
``` 

But I can do something similar to what @Bonczidai originally suggested:

```yaml 
- proxy_host:
        required: false
- proxyHost: proxy_host
```
2
------
Should I leave `overridable:false` inputs as camelCase since they aren't used outside the operation?

WDYT @oritstone @orius123  @Bonczidai @tudorlesan?

Bonczidai:
the content seems fine,
the circleci and travis/pr checks pass, but the travis/push fails with this message:
```
$ ./run 'docker run --privileged -d -p 4444 -p 49153:22 -e PORT=4444 --name docker_host_ssh orius123/dind-ssh && ./cslang-builder/bin/cslang-builder -ts ${SUITE},\!default -cov'
/home/travis/build.sh: line 41: ./run: No such file or directory
The command "./run 'docker run --privileged -d -p 4444 -p 49153:22 -e PORT=4444 --name docker_host_ssh orius123/dind-ssh && ./cslang-builder/bin/cslang-builder -ts ${SUITE},\!default -cov'" exited with 127.
```

since I am not familiar with this issue - I dare to put this one on you @orius123  :innocent: , if this failure is acceptable, feel free to merge

orius123:
Ok. Still not merging because I had issues when tried to run tests locally. 
Regarding the binding camel case from snake case. 
If I had an input that is not required and had no expression at runtime we would ignore it, but if it has an expression and you don't pass it, it would fail on name error. Sounds problematic to me. WDYT?

oritstone:
We need to find a general solution for it - not only for the marathon folder, but throughout (for example, also for ssh_command)
For now, postponing after 0.8 release

oritstone:
If an input is not required, we cannot do something like:
```yaml
- proxy_host:
        required: false
        default: proxyHost
```
Since it fails for NameError if we do not pass a value for `proxy_host` or `proxyHost`

And we cannot do something like:
```yaml
- proxy_host:
        required: false
- proxyHost: proxy_host
```
Since "Input with name: `proxyHost` is Required, but value is empty"

@SamMarkowitz @meirwah @Bonczidai @tudorlesan Do you have any idea how to solve it?

orius123:
We can use the get method I suggested.

orius123:
#420

Bonczidai:
I like the `get` approach .. another possibility which works currently is:

```yaml
    - input_name:
        required: false
    - inputName: 
        default: input_name
        required: false
```

oritstone:
In case of non-required inputs, we can do something similar to what @Bonczidai suggested:
```yaml
    - input_name:
        required: false
    - inputName: 
        default: input_name
        required: false
        overridable: false
```
where the camel case input is not overridable and not documented.
And it should work.
The only issue is it a too cumbersome syntax...
@orius123 @meirwah @Bonczidai @tudorlesan @SamMarkowitz @mihaitusa @gituser173 WDYT?

orius123:
Its too cumbersome

content/io/cloudslang/marathon/setup_marathon_core_os.sl
```diff
@@ -0,0 +1,139 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#######################################################################################################################
+# Sets up a simple marathon infrastructure on one core os host.
```
Bonczidai: core os -> CoreOS
marathon -> Marathon

:)

test/io/cloudslang/marathon/test_create_app.sl
```diff
@@ -23,8 +24,33 @@ flow:
         required: false
     - json_file
     - created_app_id
+    - is_core_os
 
   workflow:
+    - check_is_core_os:
```
Bonczidai: please extract `check_is_core_os - setup_marathon_core_os - setup_marathon` to be a standalone flow (for DRY)
maybe use `content/io/cloudslang/base/utils/is_true.sl` instead of `string_equals`
would structure the flows sthg like this:
```
setup_marathon:
  use coreos?
  setup_marathon_docker_host
  setup_marathon_coreos_os
```

test/io/cloudslang/marathon/test_create_app.sl
```diff
@@ -43,7 +69,7 @@ flow:
     - wait_for_marathon_startup:
         do:
           utils.sleep:
-              - seconds: 20
+              - seconds: 30
```
Bonczidai: using explicit sleeps causes a lot of problems in CI - please try to replace it with polling (check - sleep - check -sleep - check - timeout?)

e.g. 
https://github.com/CloudSlang/cloud-slang-content/blob/swarm/content/io/cloudslang/docker/swarm/create_cluster_with_nodes.sl
https://github.com/CloudSlang/cloud-slang-content/blob/master/content/io/cloudslang/cloud_provider/digital_ocean/v2/utils/wait_for_status_to_change.sl

Bonczidai: check if you can use `content/io/cloudslang/base/network/verify_app_is_up.sl` in cases where you waiut for the app to deploy

Bonczidai:
:+1: -- please handle comments

this PR is also fixing #347 and #376 (if so, please update PR header)

gituser173:
Done, @Bonczidai please review

Bonczidai:
please unignore marathon tests ( #376 ) to see they work

please check why the build fails (it may be the same as in [this](https://github.com/CloudSlang/cloud-slang-content/commit/47a0a9625f8a211ec2f9b7f544fa218997e4ff06) case)

gituser173:
@Bonczidai please review

content/io/cloudslang/base/network/verify_app_is_up.sl
```diff
@@ -46,12 +46,12 @@ operation:
       return_result = False
       while (( count < int(attempts) ) and ( not return_result )):
         try:
-          result = urllib2.urlopen(url)
-        except Exception :
+          result = requests.get(url, timeout=10)
```
Bonczidai: if you are staying with these changes, maybe refactor the flow to have this value as an input - WDYT?

content/io/cloudslang/base/network/verify_url_is_accessible.sl
```diff
@@ -0,0 +1,73 @@
+#   (c) Copyright 2014 Hewlett-Packard Development Company, L.P.
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Checks if an an url is accessible.
```
Bonczidai: typo: an an -> a

content/io/cloudslang/base/network/verify_url_is_accessible.sl
```diff
@@ -0,0 +1,73 @@
...
+#
+# Inputs:
+#   - url - the url
```
Bonczidai: URL for what?

content/io/cloudslang/base/network/verify_url_is_accessible.sl
```diff
@@ -0,0 +1,73 @@
...
+            - url
+            - content_type
+            - connect_timeout: "'20'"
+        navigate:
+          SUCCESS: SUCCESS
+          FAILURE: check_if_timed_out
+
+    - check_if_timed_out:
+         do:
+            base_strings.string_equals:
```
Bonczidai: maybe would be safer to use comparison here (<=) - WDYT?

content/io/cloudslang/base/network/verify_url_is_accessible.sl
```diff
@@ -0,0 +1,73 @@
...
+              - seconds: time_to_sleep
+        publish:
+          - attempts: self['attempts'] - 1
+        navigate:
+          SUCCESS: http_get
+
+  outputs:
+    - output_message: "'Url is not accessible'"
+
+  results:
```
Bonczidai: no need for explicit results

Bonczidai:
:+1: 

be sure to have 3 passing build in a row (or at least be sure that the failures are not related to marathon)

gituser173:
@Bonczidai please review the changes

content/io/cloudslang/base/network/verify_app_is_up.sl
```diff
@@ -33,9 +33,12 @@ operation:
     - time_to_sleep:
         default: 1
         required: false
+    - attempt_timeout:
```
Bonczidai: did you add this in description section?

gituser173: done

PR #257, opened 2015-07-20
removed folder name from file names

### LINKED ISSUE #49: Comments from proofing docker folder
+ general
    + file naming: inconsistent as to whether folder name is assumed or not (e.g. get_all_images vs. clear_docker_images)
    + optional parameter with `default`: inconsistent as to whether `required: false` is also present
    + some flows do not have explicitly declared and documented results 
+ docker
    + images
        + unlike other folders, contains operations and flows with the same name except flow has the word "flow" appended 
    + maintenance: should this be renamed to "mysql"
        + diskspace_health_check : should this be in the linux folder?
        + check_mysql_is_up: shouldn't `command` be `overridable: false`

SamMarkowitz:
fixes #49


SamMarkowitz:
fixes #49

SamMarkowitz:
Can't quite figure out why some of the tests are failing. Looks like there is a problem with the test flow `clear_containers` flow. (Side point: seems like this flow can be moved from the tests to the content folder.)

orius123:
We decided to go back to clear host. Haven't we? @oritstone

oritstone:
We moved to `clear_host` in the test flows in cases where it solved the disk space problem.
There are still usages of `clear_docker_containers` flow in some test flows
We can consider changing all of them back to  `clear_host`

PR #258, opened 2015-07-20
added exitStatus as an output to ssh_command.sl and ssh_flow.sl 

Alex-Persa:


content/io/cloudslang/base/remote_command_execution/ssh/ssh_command.sl
```diff
@@ -27,6 +27,8 @@
 #    - standard_out - STDOUT of the machine in case of successful request, null otherwise
 #    - standard_err - STDERR of the machine in case of successful request, null otherwise
 #    - exception - contains the stack trace in case of an exception
+#    - exitStatus - The exit status of the remote command corresponding to the SSH channel. The exit status is only available for certain types of channels, and only after the channel was closed (more exactly, just before the channel is closed).
```
oritstone: I think that a better output name would be: command_return_code
Anyway, we try to have a convention of lower case names

oritstone:
Closed by #269

PR #274, opened 2015-07-31
Feature/git reset flow (including tests)

lirantal:
Adding a new Git flow content - git_reset which resets a git working directory to HEAD or given SHA.

1. adding a git reset command to the git content flow
2. updating the git test flow with the new git reset command which will reset the cloned repository to a specific SHA hash

### DIFF
--- a/content/io/cloudslang/git/git_reset.sl
+++ b/content/io/cloudslang/git/git_reset.sl
@@ -52,7 +52,7 @@ flow:
         required: false
   
   workflow:
-    - git_clone:
+    - git_reset:
         do:
           ssh.ssh_flow:
             - host
--- a/content/io/cloudslang/git/git_reset.sl
+++ b/content/io/cloudslang/git/git_reset.sl
@@ -1,0 +1,83 @@
+# (c) Copyright 2015 Liran Tal
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+# The Apache License is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# This flow performs a git reset on a git directory to clean it up
+#
+#   Inputs:
+#       - host - hostname or IP address
+#       - port - optional - port number for running the command - Default: 22
+#       - username - username to connect as
+#       - password - password of user
+#       - sudo_user - true or false, whether the command should execute using sudo
+#       - git_reset_target - the SHA you want to reset the branch to - Defaults: HEAD
+#       - git_repository_localdir - the target directory where a git repository exists and git_branch should be checked out to - Default: /tmp/repo.git
+#       - privateKeyFile - the absolute path to the private key file
+#
+# Results:
+#  SUCCESS: git repository successfully cleaned up using git reset
+#  FAILURE: an error when trying to clone a git repository
+#
+####################################################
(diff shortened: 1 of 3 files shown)


lirantal:
When should I expect this PR to be merged?
The next PRs will be related to it because I will continue to update the test_git_flow so I'd rather not have conflicts when we try to merge it.

Not the end of the world so I don't want to rush anyone. I'll work on submitting more git related PRs soon regardless and worst-case I'll resolve the conflicts where needed.

meirwah:
@Bonczidai can you please review this?

content/io/cloudslang/git/git_reset.sl
```diff
@@ -0,0 +1,83 @@
...
+#
+# The Apache License is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# This flow performs a git reset on a git directory to clean it up
+#
+#   Inputs:
+#       - host - hostname or IP address
+#       - port - optional - port number for running the command - Default: 22
```
Bonczidai: since the default value is not explicitly declared in this flow, IMHO you should not specify it here (`Default: 22`)

test/io/cloudslang/git/test_git_flow.sl
```diff
@@ -48,11 +49,26 @@ flow:
...
 
+    - reset_git_branch:
+        do:
+          git.git_reset:
+            - host
+            - port
+            - username
+            - password
+            - git_reset_target
+            - git_repository_localdir: git_repository_localdir
```
Bonczidai: the input expression is redundant, you can write here: `- git_repository_localdir`

test/io/cloudslang/git/test_git_flow.sl
```diff
@@ -48,11 +49,26 @@ flow:
             - git_branch
             - git_repository_localdir: git_repository_localdir
```
Bonczidai: the input expression is redundant, you can write here: `-git_repository_localdir`

Bonczidai:
@lirantal , added my comments :wink:

content/io/cloudslang/git/git_reset.sl
lirantal: I'll just add a default setting value to 22, it makes sense

test/io/cloudslang/git/test_git_flow.sl
lirantal: yeah thanks for noticing that, I think it's a left-over from my previous test flow

test/io/cloudslang/git/test_git_flow.sl
lirantal: thanks

lirantal:
awesome @Bonczidai I'll address all of that tomorrow.

content/io/cloudslang/git/git_reset.sl
Bonczidai: I meant here, that it is better to remove the default value from the description - since it is not declared in the flow (which is correct - since `port` has a default value in `ssh_flow` and you don't necessary know here that value  - this is our convention, we are trying to declare default values only once, then use `required` false in the parent flows)

so just modify the input description to be `port - optional - port number for running the command` :+1:

lirantal:
thanks @Bonczidai I addressed all of your comments!

Issue #230, opened 2015-07-06
Add CI Matrix for various envs


oritstone:
Need to add a matrix to Travis\Circle CI for running content tests continuously on:
- Openstack
- Marathon
- CoreOS
- Swarm

meirwah:
we should look at these for OpenStack :
https://github.com/slimjim2234/docker-devstack
https://hub.docker.com/r/pataquets/openstack-dashboard/builds/

oritstone:
Need to be able to run automatically the following areas\tests:
- [ ] Openstack
- [ ] Stackato
- [ ] Openshift
- [ ] HP Cloud
- [ ] Chef
- [ ] Test suite: `git-local`
- [ ] Test suite: `mail-local`
- [ ] Test suite: `linux-local`
- [ ] Test suite: `amazon-local`
- [ ] Test suite: `mysql-local`
- [ ] Test suite: `digital_ocean_local`
- [ ] Test suite: `docker_local`
- [ ] Test suite: `marathon-ignore`
- [ ] Test suite: `powershell-ignore`

Bonczidai:
low coverage, also need to automate the following tests:

- [ ] 08:57:36 [INFO] - io.cloudslang.base.files.create_folder_tree
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.folders.change_ownership
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.folders.change_permissions
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.folders.create_symlink
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.folders.download_content
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.folders.make_new_folder
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.samples.deploy_tomcat_on_ubuntu
- [ ] 08:57:36 [INFO] - io.cloudslang.base.os.linux.samples.install_java_on_ubuntu
- [ ] 08:57:36 [INFO] - io.cloudslang.operations_orchestration.createCP.copy_config_items
- [ ] 08:57:36 [INFO] - io.cloudslang.operations_orchestration.createCP.create_package
- [ ] 08:57:36 [INFO] - io.cloudslang.operations_orchestration.get_central_version
- [ ] 08:57:36 [INFO] - io.cloudslang.paas.heroku.account.update_account_email
[...]

PR #281, opened 2015-08-05
Build env - swarm

Bonczidai:
branch to continue the work on swarm use case for build env (#230)

Signed-off-by: Bonczidai Levente <levente.bonczidai@hp.com>

PR #288, opened 2015-08-10
Build env - marathon

Bonczidai:
tests for Marathon for #230 


Bonczidai:
again, dependent on #281

orius123:
:100: 
Will merge after swarm ( #281 )

We need to move it to another node once Openstack is operatble

PR #283, opened 2015-08-06
OpenStack Volume and Keypair content

umavanth:
Signed-off-by: Umamaheswari S uma-maheswaris@hp.com

PR #296, opened 2015-08-13
Openstack volumes and keypair

orius123:


tudorlesan:
#283 includes all this content so this PR can be closed.

PR #299, opened 2015-08-13
Adding git_push flow

lirantal:


content/io/cloudslang/git/git_push.sl
```diff
@@ -0,0 +1,87 @@
...
+#
+####################################################
+# This flow performs a git push command to send a branch to a remote git repository
+#
+#   Inputs:
+#       - host - hostname or IP address
+#       - port - optional - port number for running the command
+#       - username - username to connect as
+#       - password - password of user
+#       - sudo_user - true or false, whether the command should execute using sudo
```
Bonczidai: missing ` - Default: false` in the description

Bonczidai:
good so far - need to add test :sunglasses:

lirantal:
@Bonczidai updated - I think we also need to update the other git flows for that default  for sudo_user field.

What kind of test can I add? pushing to a repository requires permissions and I'm not sure how it would be easy to add test but I'm all ears if you have some kind of flow to suggest.

Bonczidai:
I see here two possibilities:
- we could set up a Docker container with Git server on it - this could be used for all the Git flows
- add local test with `git-local` test suite

We can go with the first one which is a bit more complex or go with the second one and do the automation later - @meirwah WDYT on this?

lirantal:
@Bonczidai both options sound somewhat complex to me so I would love to get your help on this if we want to go for any of them.

Bonczidai:
sure - I'll be happy to guide you once we decide which option to go with

meirwah:
let's start with "git-local" , sanity test, and add later complex (using docker) tests...

lirantal:
I don't know what's "git-local", so whose taking this further?

meirwah:
@orius123  can work on adding a sanity test...

PR #309, opened 2015-08-21
Add agent to swarm tests

### LINKED ISSUE #286: Add node to swarm cluster used in tests
now the swarm tests use a cluster of one machine (manager + 1 node - two CoreOS machines) - we should use the third CoreOS machine as a node

this should not be a hard task (most of the work is around modifying the test flows) - ideal task for those who want familiarize with swarm content

Bonczidai:
fixes #286 


test/io/cloudslang/docker/swarm/demo_containers.inputs.yaml
```diff
@@ -18,5 +18,6 @@ test_demo_containers_success:
     - host: <coreos_host_1>
     - username: core
     - private_key_file: <private_key_file>
-    - agent_machine_ip: <coreos_host_2>
+    - agent_machine_ip_1: <coreos_host_2>
```
meirwah: it's not better to have here a list?

Bonczidai: right -- and we can use async loop when doing independent agent machine operations

Bonczidai:
addressed the issue: used lists for agent data and async loop where possible (reduce build time with 1 min)

meirwah:
:+1:

Issue #357, opened 2015-09-24
Math comparisons ops [5]


meirwah:

oritstone:
- Is Equal
- Not Equal
- Greater Than
- Greater Than or Equal
- Less Than
- Less Than or Equal

SamMarkowitz:
What kind of results do we expect from these? 

1. `TRUE` and `FALSE` as you would get in a programming language 
2. Or something like `EQUALS` and `NOT_EQUALS`. 

If option 2, doesn't it make sense to have fewer operations. For example it would be strange to have both an `equals` and `not equals` operation that would both result in either `EQUALS` or `NOT_EQUALS`?

@oritstone, @Bonczidai

Bonczidai:
I would go with the second approach similar to what we have in DO content -> [compare_float](https://github.com/CloudSlang/cloud-slang-content/blob/9ada768d6a5af3abaabf0a831adb26540106e439/content/io/cloudslang/base/math/comparisons/compare_float.sl) - maybe enhance it and add an `equals` op -- I don't like adding lots of operations like <, <=, >, >= since we easily encapsulate these checks in task navigation section and the additional computing effort in worst case is 1 check in the result condition:

```
  navigate:
    GREATER_THAN: >
    EQUALS:  ==
    LESS_THAN: <
```

lets hear what others think - adding @mihaitusa

PR #370, opened 2015-09-30
Handle zombie droplets - DigitalOcean content

### LINKED ISSUE #285: Sanity check for build env droplets
make sure there are no zombie droplets

Bonczidai:
fixes #285 

Signed-off-by: Bonczidai Levente levente.bonczidai@hpe.com


Bonczidai:
@mihaitusa please pre-review -- tests need to be added

mihaitusa:
:+1:  @Bonczidai  ... Only thing... check up on first Travis build.

Bonczidai:
Some notes:

- http call's `keepAlive` changed to `false` ( CloudSlang/score-actions#80 )
- added compare_float op (related to #357 )
- instead of waiting an explicit amount of time, I added a a flow (wait_for_status_to_change) based on polling.. the logic becomes more complicated but the flow is more stable and the execution time is reduced --> I think we should all apply this convention instead of `sleep` calls with explicit values

@meirwah @oritstone @SamMarkowitz @mihaitusa @gituser173

Bonczidai:
this content has test which costs 1 additional droplet / build

oritstone:
@gituser173 please review

gituser173:
Looks good :+1:

PR #372, opened 2015-10-04
Adding google-api-python-client

mbettan:
Adding google-api-python-client library as I will publish Google Container Engine depending on it


mbettan:
Duplicate with other pull request.

PR #374, opened 2015-10-06
Additional content for HP Cloud and Chef [2]

benc-uk:
## Overview
This pull request includes new content for HP Cloud and Chef, in addition a small number of supporting utility operations are included.

Chef content is located in namespace **io.cloudslang.chef**
HP Cloud is located in namespace **io.cloudslang.cloud_provider.hp_cloud**

### Pre-reqs
 - In order to use the HP Cloud content you will require an account on HP Cloud. All operations are HTTP API based. This content has been based on the OpenStack operations and flows, but modified to work with HP Cloud, in addition content for IP allocation, assignment and removal have been added
 - In order to use the Chef content you will require a functioning Chef deployment, namely a Chef Server and Linux machine configured with knife. The machine running knife and and Chef server can be the same. All Chef operations are via SSH

### Utils
io/cloudslang/base/network/wait_port_open.sl - **Poll an IP address, checking until TCP port is open**
io/cloudslang/base/strings/filter_lines.sl  - **Filter input returning only lines with matching filter**
io/cloudslang/base/utils/check_bool.sl  - **Simple boolean flow branch**
### Chef
io/cloudslang/chef/assign_cookbooks.sl - **Assign Chef cookbook to a node**
io/cloudslang/chef/bootstrap_node.sl - **Bootstrap a server, adding it to Chef**
io/cloudslang/chef/delete_node.sl - **Remove node from Chef**
io/cloudslang/chef/knife_command.sl - **Run a Chef knife command**
### HP Cloud
io/cloudslang/cloud_provider/hp_cloud/create_server.sl - **API call to create a server (op)**
io/cloudslang/cloud_provider/hp_cloud/create_server_flow.sl - **Main flow to create a server and assign IP to it**
io/cloudslang/cloud_provider/hp_cloud/delete_server.sl - **API call to delete a server (op)**
io/cloudslang/cloud_provider/hp_cloud/delete_server_flow.sl - **Main flow to terminate a server and release the IP**
io/cloudslang/cloud_provider/hp_cloud/get_authentication.sl - **API call to get auth token (op)**
io/cloudslang/cloud_provider/hp_cloud/get_authentication_flow.sl - **Main flow required to logon and get auth token**
io/cloudslang/cloud_provider/hp_cloud/get_server_details.sl - **Get server details (op)**
io/cloudslang/cloud_provider/hp_cloud/get_server_state_flow.sl - **Get the current status of the server**
io/cloudslang/cloud_provider/hp_cloud/net/add_ip_to_server.sl - **Attaches IP to a server (operation)**
io/cloudslang/cloud_provider/hp_cloud/net/create_floating_ip.sl - **API call to allocate a new floating IP (op)**
io/cloudslang/cloud_provider/hp_cloud/net/create_floating_ip_flow.sl  - **Flow to allocate a new floating IP**
io/cloudslang/cloud_provider/hp_cloud/net/delete_floating_ip.sl - **API call to delete a floating IP (op)**
io/cloudslang/cloud_provider/hp_cloud/net/delete_floating_ip_flow.sl - **Flow to delete a floating IP**
io/cloudslang/cloud_provider/hp_cloud/net/list_all_floatingips.sl - **API call returns all floating IPs (ops)*
io/cloudslang/cloud_provider/hp_cloud/net/list_all_networks.sl - **API call returns all networks (op)**
io/cloudslang/cloud_provider/hp_cloud/utils/get_external_net_id.sl - **Util operation to parse JSON and get external network from list**
io/cloudslang/cloud_provider/hp_cloud/utils/get_floatingip_id.sl - **Util operation to parse JSON and get the id for a floating IP**

## Usage and Tests
Test flows are included in test/io/cloudslang in the relevant sub directories. Example testsuite input files are also provided, however as most inputs for HP Cloud and Chef are environment specific they are just placeholders.

 - test/io/cloudslang/chef/**test_chef_bootstrap_run.sl**
  - This tests Chef bootstrap functionality on existing unmanaged server
 - test/io/cloudslang/chef/**test_chef_hpcloud_full.sl**
  - This is a full end to end test combining HP Cloud to build a new server, then uses Chef to deploy an appserver to the new server, then checks it is online and working
 - test/io/cloudslang/cloud_provider/hp_cloud/**test_hpcloud_deploy_terminate.sl**
  - Deploys a new server instance in HP cloud, then removes it. Clean test
 - test/io/cloudslang/cloud_provider/hp_cloud/**test_hpcloud_ip_allocation.sl**
  - Allocates a floating IP then releases it. Clean test


orius123:
Looks awesome!! Plus gr8 description!! 
10x for the contribution.

meirwah:
:+1:

oritstone:
Thanks a lot @benc-uk!
Really nice work!

We will get to fully review it shortly.

In the mean time, I saw that you gave 2 example flows of how to use the content.
In our contribution guide: https://github.com/CloudSlang/cloud-slang-content/blob/master/CONTRIBUTING.md, we ask for new content to be tested.
You can see guidelines to how we write tests in our documentation: http://cloudslang-docs.readthedocs.org/en/v0.8/cloudslang_tests
In general, we have a framework for writing test cases, and these test cases use test flows, very similar to the examples you supplied above.
Can you add these flows as tests to either: `test/io/cloudslang/chef` or `test/io/cloudslang/cloud_provider/hp_cloud`?
In addition, if you manage to also supply the test cases we will be very grateful  :)
We are available, of course, for any assistance on the matter.

benc-uk:
Hi Thanks @oritstone 

I looked at the test case thing, but wasn't sure about it
It looked like there was an one-to-one relation between test case the subject flows, so I'd need to create 24 test cases? My content has logical dependencies that would make it difficult to create standalone "atomic" test. I was also unsure how I could provide tests that did anything without sharing my usernames, passwords, hostnames and other environment specific details?

Maybe I've misunderstood. I'll take another look, do some reading and makes some changes over the next few days

meirwah:
Hi @benc-uk , no need for 24 test-cases, it's to much...

benc-uk:
Right. I think I've got my head round this test thing. I'll write some better test flows, ones that don't muddle the HP Cloud and Chef stuff into one huge flow
Is the name of the test suite important? I've just gone for 'hp_cloud' and 'chef' which seems logical!

oritstone:
The test suite name is used for knowing where to run the test during our CI process - in case running the test requires a specific environment (such as Chef)
You can put any test suite you want, and it will not run automatically as part of our CI process at first (only manually), until we manage to set up a suitable environment for our CI environment (we might need assistance from you later on for that part...)

Regarding the number of test cases, in most cases we have one test case cover up a a lot of the content. There is no actual need to write a test case per each flow. But it is a good idea to have separate test cases for Chef & for HP Cloud

benc-uk:
OK I've added some proper test flows and some input.yaml testsuites. See end of updated comment for more details
The Travis CI build is failing, I don't understand the reason, it's saying 'get_authentication_flow' has missing dependencies, but I'm not seeing any error with that flow on my system. Halp! :(

oritstone:
The problem is that you are not updated to our latest version of the repo. There were some major content commits since you last updated from our master...

The `parse_authentication operation` is no longer present in our content. It was replaced by a more general purpose operation called: `io.cloudslang.base.json.get_value_from_json`

content/io/cloudslang/cloud_provider/hp_cloud/get_authentication_flow.sl
```diff
@@ -0,0 +1,78 @@
...
+    - region
+    - proxy_host:
+        required: false
+    - proxy_port:
+        required: false
+
+  workflow:
+    - get_token:
+        do:
+          hpcloud.get_authentication:
```
oritstone: We already have a flow in our base content for HTTP POST: `io.cloudslang.base.network.rest.http_client_post` 
It is preferable to use this wrapper flow instead of using the HTTP Java action directly through the specific `io.cloudslang.cloud_provider.hp_cloud.get_authentication` operation.

content/io/cloudslang/base/utils/check_bool.sl
```diff
@@ -0,0 +1,29 @@
...
+#   - check_bool - Boolean value to check
+# Results:
+#   - SUCCESS - check_bool is true
+#   - FAILURE - check_bool is false
+####################################################
+
+namespace: io.cloudslang.base.utils
+
+operation:
+  name: check_bool
```
oritstone: I think that a better name for the operation might be *is_true*   :)
WDYT?

content/io/cloudslang/base/utils/check_bool.sl
```diff
@@ -0,0 +1,29 @@
...
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Check is boolean is true or false, use for flow control
+#
+# Inputs:
+#   - check_bool - Boolean value to check
```
oritstone: The name of the input is `bool_value`

content/io/cloudslang/chef/assign_cookbooks.sl
```diff
@@ -0,0 +1,65 @@
...
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Assigns one or more Chef cookbooks (comma seperated) to a node's run list
+#
+# Inputs:
```
oritstone: Missing doc about `cookbooks` & `node_name`

content/io/cloudslang/chef/knife_command.sl
```diff
@@ -0,0 +1,72 @@
...
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Run Chef knife command and return filtered result
+#
+# Inputs:
+#   - knife_host - Server with configured knife accessable via SSH, can be main Chef server
+#   - knife_username - SSH username to access server with knife
+#   - knife_password - optional - If using password auth
+#   - knife_privkey - SSH keyfile (local file that resides where flow is executing)
```
oritstone: add `- optional -`

content/io/cloudslang/chef/knife_command.sl
```diff
@@ -0,0 +1,72 @@
...
+#   - knife_cmd - Knife command to run e.g. 'cookbook list'
```
oritstone: Missing doc about input *knife_timeout*

content/io/cloudslang/chef/knife_command.sl
```diff
@@ -0,0 +1,72 @@
...
+        required: false
+        
+  workflow:
+    - knife_cmd:
+        do:
+          ssh.ssh_command:
+            - host: knife_host
+            - username: knife_username
+            - password: 
+                required: false
```
oritstone: We recently changed the syntax of task arguments. It is not like inputs now, but rather like method invocation arguments. Meaning we do not support properties for task arguments besides assigning a default value.
In the new syntax it should be:
```yaml
- password: knife_password
- privateKeyFile: knife_privkey
```

benc-uk:
Hi @oritstone I've committed a few changes in line with your comments and mistakes you found.

content/io/cloudslang/base/network/wait_port_open.sl
```diff
@@ -0,0 +1,57 @@
...
+
+  action:
+    python_script: |
+      import socket, time, sys
+      is_open = False
+
+      for t in range(0, int(tries)):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        sock.settimeout(float(timeout))
```
Bonczidai: can we set the timeout (maximum wait time) here, so we will not need to `sleep` and `for` ?

benc-uk: I wanted the operation to act synchronously, and poll until the port was available - mainly when building a new cloud instance (or rebooting) while the VM comes up.

The sleep is actually only used when the *sock.connect_ex* connects but it's not in the fully open state yet. Mostly when connection fails the *sock.connect_ex* call will wait the normal timeout period and return null, however the network goes through a few interesting transition states, and sometimes it will return very quickly without a timeout so the sleep is for this edge case.

I wanted to keep the operation "standalone" I suppose I could remove the for loop polling and move that to a loop at the flow level. I did try that before, but this solution seemed much less ugly to me.

Bonczidai: :+1: thx for the explanation - let's leave it like this, I also like the "standalone" way

content/io/cloudslang/chef/bootstrap_node.sl
```diff
@@ -0,0 +1,99 @@
...
+            - text: returnResult
+            - filter: node_host
+        publish:
+          - filter_result
+
+  outputs:
+    - raw_result: returnResult
+    - knife_result: "standard_err  + ' ' + (filter_result if 'filter_result' in locals() else returnResult)"
+    - standard_err: standard_err
+    - node_name
```
Bonczidai: is there any advantage of declaring `node_name` as an output? I see that it's a flow input and is not modified in the flow

a situation I can think of: in a task's (which calls this flow) publish section you want to refer this input -- but in that case the proper way to do it is using our `fromInputs` construct

benc-uk: There's no real reason to have it there, It was put there as a convenience to help parent flows that call this, it can be changed to be fromInputs if that's preferred way

Bonczidai: I see there is a good use case in `test_chef_hpcloud_full.sl`: you bind `node_name` there (`server_name + '_' + ip_address`) and you want this value to be accessible in other tasks (`fromInputs` does not help here because it contains flow context, but not the task inputs -- btw we are planning to change its name to a more informative one)

benc-uk: Exactly, the node name in Chef needs to be unique, so it's common to concat the IP address or some other value, to make something sensible (human readable) but unique. I did experiment with fromInputs once but didn't fully understand why it was a good or helpful thing!

content/io/cloudslang/cloud_provider/hp_cloud/create_server.sl
```diff
@@ -0,0 +1,87 @@
...
+# Inputs:
+#   - server_name - Name for the new server
+#   - img_ref - Image id to use for the new server (operating system)
+#   - flavor_ref - Flavor id to set the new server size
+#   - keypair - Keypair used to access the new server
+#   - tenant - Tenant id obtained by get_authenication_flow
+#   - token - Auth token obtained by get_authenication_flow
+#   - region - HP Cloud region; 'a' or 'b'  (US West or US East) 
+#   - proxy_host - optional - proxy server used to access the web site - Default: none
+#   - proxy_port - optional - proxy server port - Default: none
```
Bonczidai: missing description for `network_id`

benc-uk: I copied this from **create_openstack_server** operation and I notice that it too is missing a description for `network_id`. However I'll add something

Bonczidai: gr8.. please also add the description in `io.cloudslang.cloud_provider.hp_cloud.create_server_flow` (and in other flows, if they have this input)

benc-uk: No other flows reference this input as I never use it, it's optional

Bonczidai:
besides benc-uk/cloud-slang-content#1 PR there are two issues left:

- if you look in CIrcleCI build ( https://circleci.com/gh/CloudSlang/cloud-slang-content/1997 ) at the command that runs cslang builder:

circle1

at the coverage section:

circle2

you see that one of your flows is not covered (not called in your test flows)

- your operations use http action directly -- they should be flows and use rest operations get / post / delete (this is the more convenient and preferred way):
  - rest ops: content/io/cloudslang/base/network/rest
  - example flows - openshift content: content/io/cloudslang/paas/openshift/cartridges/add_cartridge.sl

the openstack content that you looked at, will also be updated in the future to use rest ops

If you need further assistence with the issues => we are here

benc-uk:
Started task of switching all 9 operations that call `HttpClientAction` directly over to the new **http_client_post** (etc) flows 

I can not get it to work, it always complains that "username" is a required but empty. It seems to want *all the optional inputs* even though they are set to required: false.
I don't want to add another 10 inputs to all my flows just to pass through empty values. 

The OpenStack REST calls will never require a username or password so having them as optional inputs (like on the openshift flows) makes no sense.

oritstone:
I tried your example flow above and it works fine for me...
Maybe you have some other version of the *http_client_get * flow, for which the username is a required parameter.
Which version of the content are you using? 
Your content was packed with the CLI you downloaded? If so, which version of the CLI?

Bonczidai:
*quick answer*: you probably use an old CLI - please download https://github.com/CloudSlang/cloud-slang/releases/tag/cloudslang-0.8.3 (which suports the DSL changes in task syntax)

*longer answer*:
because of the old CLI which does not support new task syntax the `username` input in `http_client_get` flow's task:
```
  workflow:
    - http_client_action_get:
        do:
          http_client_action:
            - url
            - username
            - password
...
```

cannot be bound since CLI sees it a required input. Old DSL used modifiers like (`required: false`) on task inputs, in new DSL they are not required by default (and we don't support any modifier for task arguments)

benc-uk:
Thanks @Bonczidai that was exactly what was wrong!

I was using v0.8.0.1 (last stable) I've downloaded 0.8.3 and it is working. I am having to remove `required: false` at the task level on some of my flows, but that's good

benc-uk:
Switched over to the new HTTP flows. That's all much better, and neater looking :+1:

Bonczidai:
:100: awesome..

next step is for us to set up environments and run your test flows

benc-uk:
Let me know if you need any help, HP Cloud is pretty easy, just need to register. Getting a working Chef environment however takes quite a bit more work.

Bonczidai:
HP cloud is ok (benc-uk/cloud-slang-content#2), next step will be Chef

benc-uk:
This is a very good guide to setting up Chef. https://www.digitalocean.com/community/tutorials/how-to-set-up-a-chef-12-configuration-management-system-on-ubuntu-14-04-servers
You can make the Chef workstation and the Chef server the same machine 

I have a small chef repo you can clone https://github.com/benc-uk/my-chef-repo It has some standard cookbooks, which will save you downloading lots of stuff

Bonczidai:
thanks, looks like a good tutorial -- I'll give it a try

benc-uk:
:sos: Help!
I spotted I have expressions checking the status_code in the `results` part of my REST call flows, making sure it was the correct HTTP code e.g. 201. 
```
  results:
    - SUCCESS: "'status_code' in locals() and status_code == '201'"
```

This worked when they were **operations**, but now they are **flows** it is ignored, so all my error checking is being ignored! :anguished: Questions:
 - Is there anyway to control the result SUCCESS/FAILURE of a flow with an *inline* expression just like in an operation?
 - If not, what is the best way to check a status code in the flow? Use a `string_equals` step? It seems a bit clumsy and verbose.

benc-uk:
This example test flow shows how you can call http_client_get flow and even when returning a 404 error the flow will exit with SUCCESS

```
namespace: bencoleman.test

imports:
  rest: io.cloudslang.base.network.rest
  print: io.cloudslang.base.print

flow:
  name: testflow
  inputs:
  workflow:
    - simple_get:
        do:
          rest.http_client_get:
            - url: "'http://example.com/this_url_doesnt_exist'"
        publish:
          - return_result
          - status_code
          - return_code

    - print_result:
        do:
          print.print_text:
            - text: "'return_code='+ return_code + ' | status_code=' + status_code"

    - on_failure:
      - catch_errors:
          do:
            print.print_text:
              - text: "'This never is called, even with a 404 error!' + return_code" 
```

Bonczidai:
good catch @benc-uk , unfortunately currently there is no DSL support for such cases -- so you have to introduce additional task / tasks to check the values (utility operations can be found in content/io/cloudslang/base/strings folder -> equals, occurence counter)

do you think it would be worthwhile to add such support in the DSL?

p.s. we should throw error in cases when flow results have expressions, we have an issue about it ( CloudSlang/cloud-slang#327 )

also adding @oritstone

benc-uk:
I've made a small modification to `http_client_action` in this commit https://github.com/benc-uk/cloud-slang-content/commit/237c5c4d187b7e74449586863bb281a124b97375 

The operation now matches its description - I don't know if the HTTP code checking was removed for a reason? However this way it means people that call `http_client_*` do not need to implement their own error checking, but there might be some cases where a HTTP code outside the 200-299 range is considered acceptable, but I think these cases are very small

Bonczidai:
I think we can add your modification to `http_client_action` to validate the range, but you still need to check in your whether the status code is the expected one (e.g. 202)

orius123:
Maybe http op can get the expected status as input, or maybe better wrap it in a flow that dies it instead of writing this validations on many places.

benc-uk:
Good idea, implemented here https://github.com/CloudSlang/cloud-slang-content/commit/7b62d336614f2230dd2eb14669f4c366cf8db981

I think most users will never need to change the default, but the option is there now and they can provide a python list for flexibility, e.g [300, 302]

For my HP Cloud flows, I don't think I need to check the exact code returned, long as it is 2xx I am OK. The issue was, one flow had a typo in the URL causing a 404 that was never getting trapped

benc-uk:
RE https://github.com/CloudSlang/cloud-slang-content/pull/374#commitcomment-13791150 @Bonczidai  :blush: Yes! That catches me out all the time in Python

With the input name, yes, all lower case (snake_case) is the convention used everywhere else, but this operation has **LOTS** of inputs which for some reason are all camelCase. I didn't want my new input to be different. Maybe refactoring the input names can be an exercise left to another time?

Bonczidai:
**ok** - let's leave camel case for now.. they are camel case because we are investigating what is the best way to make operation inputs ( that use Java actions ) snake case (it is problematic because under the hood the Java method inputs are bound based on these names)

just be sure - if you reference this input in other flows - to use: `- validHttpStatusCodes: valid_http_status_codes` assigment

Bonczidai:
@benc-uk please merge from master to resolve conflicts

test/io/cloudslang/chef/test_chef_bootstrap_run.inputs.yaml
```diff
@@ -0,0 +1,25 @@
...
+#   All rights reserved. This program and the accompanying materials
+#   are made available under the terms of the Apache License v2.0 which accompany this distribution.
+#
+#   The Apache License is available at
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+test_chef_bootstrap_run_success:
+  testSuites: [chef-local]
+  description: Tests that HP Cloud and Chef works
```
Bonczidai: please remove `HP Cloud` from description -- this test only calls chef flows :cop:

Bonczidai:
hi @benc-uk -- I have an issue running the `test_chef_hpcloud_full` test:

the `chef_run_client` task (cmd: `sudo chef-client`) fails with message:
```
[2015-10-19T12:06:38+00:00] INFO: HTTP Request Returned 404 Not Found: 
[2015-10-19T12:06:40+00:00] INFO: HTTP Request Returned 412 Precondition Failed: {"message"=>"Run list contains invalid items: no such cookbooks apt, apache2.", "non_existent_cookbooks"=>["apt", "apache2"], "cookbooks_with_no_versions"=>[]}
```

however `apt` and `apache2` cookbooks (from your repo) are present in my `chef-repo/cookbooks` folder:
```
root@Chef:~# ls -l chef-repo/cookbooks/
total 12
drwxr-xr-x 7 root root 4096 Oct 16 09:27 apache2
drwxr-xr-x 9 root root 4096 Oct 14 10:03 apt
-rw-r--r-- 1 root root 3064 Oct 15 09:19 README.md
```

Any idea why?

benc-uk:
Could be two things. 
- I think you haven't uploaded the cookbooks to the chef-server, run `knife cookbook list` to list all the cookbooks the server has. Having cookbooks in your chef-repo doesn't mean the server has them, if they are missing you will need to run `knife cookbook upload apt` (and apache2)

- Officially the proper syntax for adding to a run_list would be like `"recipe[apt], recipe[apache2]"`, however the `knife node run_list add` lets you shortcut it with just `"apt, apache2"` (the recipe part is assumed). Note, you can also add what chef calls "roles", e.g. `"role[webserver]"` but lets leave that for now. **NOTE.** Weirdly the `run_list remove` command does **not** assume the same shortcut! So it is best to write your run_list out in full, e.g. `"recipe[apt], recipe[apache2]"`

Bonczidai:
thanks.. it was the first one

I am still having issues with the last step ( `net.verify_app_is_up` ), it cannot access the machine through the specified port (I am using Tomcat cookbook) -- but I guess this is a configuration issue related to Tomcat (when I ssh into the machine in HP Cloud Tomcat is installed correctly)

do you plan any additional changes on this branch? (is it ready for merge?)

benc-uk:
re: Checking Tomcat is "up", it could be the port number check it is 8080, otherwise it might be the firewall on HP Cloud, go into Compute -> Access & Security -> Security Groups and edit the rule for the *default* group, and allow 8080 through. You should be able to easily check with your browser

I'm not planning any more changes on this branch, I think it's ready for merge

PR #412, opened 2015-11-10
Nutanix - content corrections (beta)

mihaitusa:
Added several beta operations.


content/io/cloudslang/nutanix/beta_tasks_poll.sl
```diff
@@ -0,0 +1,72 @@
...
+    - proxy_username:
+        required: false
+    - proxy_password:
+        required: false
+    - taskId
+
+  workflow:
+    - delete_vm:
+        do:
+          rest.http_client_get:
```
mihaitusa: @mbettan  You sure about this (use of http_client_get for a delete step)?

content/io/cloudslang/nutanix/beta_vms_clone.sl
```diff
@@ -0,0 +1,78 @@
...
+        required: false
+    - proxy_port:
+        default: "'8080'"
+        required: false
+    - proxy_username:
+        required: false
+    - proxy_password:
+        required: false
+    - templateId
+    - body
```
mihaitusa: @mbettan  I think the body of a Post API call should not be present as input but constructed inside the http_client_post step.

content/io/cloudslang/nutanix/beta_vms_create.sl
```diff
@@ -0,0 +1,78 @@
...
+        required: false
+    - proxy_port:
+        default: "'8080'"
+        required: false
+    - proxy_username:
+        required: false
+    - proxy_password:
+        required: false
+    - templateId
+    - body
```
mihaitusa: @mbettan  Same comment as above.

mihaitusa:
@gituser173 please review!

test/io/cloudslang/nutanix/test_nutanix_complete.sl
```diff
@@ -0,0 +1,127 @@
...
+        publish:
+          - return_result
+          - response
+          - error_message
+          - response_body: return_result
+
+    - print_reateResourceVMClone:
+        do:
+          print.print_text:
+            - text: response
```
gituser173: assert here? and in the other tests.

test/io/cloudslang/nutanix/test_nutanix_complete.sl
```diff
@@ -0,0 +1,127 @@
...
+        publish:
+          - return_result
+          - response
+          - error_message
+          - response_body: return_result
+
+    - print_nutanixCreateVM:
+        do:
+          print.print_text:
+            - text: response
```
gituser173: same comment as before

content/io/cloudslang/nutanix/beta_create_resource_vmclonedto.sl
```diff
@@ -0,0 +1,70 @@
...
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+####################################################
+# Creates an create.dto.acropolis.VMCloneDTO object.
+# For example, it could be used for clone a Virtual Machine operation.
+#
+# Inputs:
+#   - name - name for the cloned Virtual Machine
+#   - numVcpus - overrides the number of vCPUs assigned to the clone
+#   - memoryMb - overrides the amount of RAM assigned to the clone
```
gituser173: snake case maybe? WDYT?

oritstone:
We need to find a way to be able to test this content
For now leaving it on hold

mbettan:
It would hard as Nutanix is hyperconverged infrastructure. Already tried to find a simulated API. Only way would be to work with Nutanix or their partner to get physical hardware. I had tested those operations through HP OO calls in customer site. Don't have access anymore to validate again.

oritstone:
It is problematic for us to not be able to test it, since, for example, we just made some changes to our language structure (https://github.com/CloudSlang/cloud-slang/pull/476), so we already cannot verify that the content still works after changes like this...
We will continue to try to find a way to test that content, so we can add it to our repository.
If you happen to have access to such environment, or have an idea how it can be done, we will appreciate it a lot if you can assist in that.

mbettan:
I will contact my customer to check if we can test remotely with him, but it will a one shoot test, we won't get a permanent access. What do you think? Who could be available with me for this remote call?

By the way, are we using in background some Service Virtualization technology? If we can capture those API calls once and can reproduce offsite that would be very efficient to validate clouds lang content.

mbettan:
I think we will get more and more content without endpoint access for testing due to hard requirements or paying access. Moreover, any new release can break it. We really need to find a way to capture and simulate it offsite.

mbettan:
Nutanix Prism Community Edition seams available, otherwise we will have to test the content with a partner/customer.

mbettan:
Can we publish them as beta version?

tethryus:
Pull Request closed until a better implementation with Nutanix will be available.

PR #421, opened 2015-11-16
Fix git content according to the new expressions and string constants syntax 

oritstone:


Bonczidai:
into `master`? or `change_double_quotes_and_expressions_syntax`

oritstone:
It is above the `change_double_quotes_and_expressions_syntax` as all the other PRs that deal with the changes to the DSL.
We should review it, and **not** merge anything until we merge the cloud-slang PR: https://github.com/CloudSlang/cloud-slang/pull/476
Before merge we also need to test all the content that does not have automatic testing

PR #429, opened 2015-11-18
Align jenkins content to the new expressions and string constants syntax

oritstone:


mihaitusa:
@oritstone Everything looks ok in here to.

PR #430, opened 2015-11-18
Fix git content according to the new expressions and string constants syntax

oritstone:


mihaitusa:
@oritstone  Everything looks ok. I will merge it.

PR #460, opened 2015-11-25
docker local

### LINKED ISSUE #307: demo_dev_ops network issue on 0.8
I tested demo_dev_ops.sl on Windows on different HP networks and on one of our remote Docker machines and in each case the last step (test_application) times out, possibly because of a proxy. The application is actually up and can be checked from the browser though. 

Also, because this step fails, send_error_mail step should be executed, which concatenates a string with the output of test_application which was renamed from error_message to output_message so then the flow throws a Slang Error. I already fixed this though it's not merged to 0.8 but to master.

gituser173:
Fixes #307 
Fix demo_dev_ops network issue

### DIFF
--- a/content/io/cloudslang/docker/containers/examples/demo_dev_ops.sl
+++ b/content/io/cloudslang/docker/containers/examples/demo_dev_ops.sl
@@ -104,10 +104,10 @@ flow:
 
     - test_application:
         do:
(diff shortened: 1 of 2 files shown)


oritstone:
@gituser173 is it ready for review?

mihaitusa:
:+1:  Looks ok!

PR #472, opened 2015-12-08
Heroku

FlorianJDF:
Hello everybody,

I've done some operations to manage a Heroku account.
They use the Heroku API (api.heroku.com) thanks to the http_client_action.

All the operations are in the folder pass/heroku.
For each operation I've created a test case.

<b>How to test :</b>
The tests for each Heroku operation can be found in the test/ repository.
They are sorted with the same folders and names as the operations.
They can be launched by using the .sl files with command line, or by using the Build Tool, because a .input.yaml file exists for each test case.

-- 
Florian TEISSEDRE
florian.teissedre@hpe.com


content/io/cloudslang/paas/heroku/SSH/list_account_ssh_keys.sl
```diff
@@ -0,0 +1,46 @@
...
+  inputs:
+    - username
+    - password
+  workflow:
+    - account_ssh_keys:
+        do:
+          imports_operations.http_client_action:
+            - url: 'https://api.heroku.com/account/keys'
+            - method: "GET"
+            - username: ${username}
```
mihaitusa: not needed ${username}

content/io/cloudslang/paas/heroku/SSH/list_account_ssh_keys.sl
```diff
@@ -0,0 +1,46 @@
...
+            - password: ${password}
+            - contentType: "application/json"
+            - headers: "Accept:application/vnd.heroku+json; version=3"
+            - validHttpStatusCodes: ${ range(200, 600) }
```
mihaitusa: Why the valid range is between 200 and 600? ... usually we use this kind of list only for successful statusCode's range(200, 300) which means that the list is 200 ... 299.

content/io/cloudslang/paas/heroku/account/current/update_account.sl
```diff
@@ -0,0 +1,59 @@
...
+flow:
+  name: update_account
+  inputs:
+    - username
+    - password
+    - json_body
+  workflow:
+    - update_account:
+        do:
+          imports_operations.http_client_action:
```
mihaitusa: @FlorianJDF  Here isn't suppose to be: imports_operations.http_client_post or imports_operations.http_client_put for example?

content/io/cloudslang/paas/heroku/account/current/details_account.sl
```diff
@@ -0,0 +1,48 @@
...
+
+flow:
+  name: details_account
+  inputs:
+    - username
+    - password
+  workflow:
+    - details_account:
+        do:
+          imports_operations.http_client_action:
```
mihaitusa: @FlorianJDF  Here is better to be: imports_operations.http_client_get That's why we did the lightweight versions: http_client_get, http_client_delete, http_client_post, http_client_put so we don't need to specify all the inputs from http_client_action

content/io/cloudslang/paas/heroku/SSH/list_account_ssh_keys.sl
FlorianJDF: @mihaitusa I put the range between 200 and 600, because with the version 0,9, the operation http_client_action goes into FAILURE if the http output is higher than 400.
But I wanted to analyse the output of the HTTP action. I choose to change the range instead of using the "navigate" on the FAILURE.

content/io/cloudslang/paas/heroku/account/current/details_account.sl
FlorianJDF: @mihaitusa If I use the http_client_get and other lightweight operations, I'm not able to edit the "validHttpStatusCodes".

PR #479, opened 2015-12-29
Flow (and ops)  - Create an OO Central Content Pack

talsch:
I have created a flow which packs a cslang folder into a deployable content pack.
I could do it in one python script but I decided it will be nicer to do it with a flow (and cslang is light enough), using existing operations.

Tested it on my windows machine.
Please review (please test on Linux. I tested on Windows).

It contains:
create_package – This is the “master flow” which creates the cp.
copy_config_items – a Python operation which packs the configuration items. (It iterates  in a folder tree. Can be reused maybe…).
Create_folder_tree – an operation which is a variant of the create folder operation from the base. 
The difference is that it uses the python os.makedirs(folder_name) instead of   os.mkdir(folder_name) so it supports creating the entire path to the folder if it does not exist.
My opinion is that this should be fixed in the base.

I did not add an operation for deployment on Central, but we can add a navigation to deployment (REST put…) so it will also deploy in Central  in one click.
WDYT?

Regards,
Tal.


mbettan:
Great content ! It would be good to add an operation to deploy also automatically in central I think.

We will integrate your workflows in the CloudSlang POC appliance we are building for demonstrating CloudSlang on customer site and demos environments with @FlorianJDF

Bonczidai:
@talsch you can also open an issue for the `os.makedirs(folder_name)` enhancement, so we'll be able to tackle it (or you can do it and contribute back)

mihaitusa:
@talsch  IMHO we should create "operations_orchestration" folder and put "create_cp" as a subfolder in it. Also the "create_folder_tree.sl" and "copy_config_items.sl" would be better to stay under "operations_orchestration/utils".

content/io/cloudslang/createCP/create_package.sl
```diff
@@ -0,0 +1,124 @@
...
+####################################################
+namespace: io.cloudslang.createCP
+
+imports:
+ base_files: io.cloudslang.base.files
+
+flow:
+  name: create_package
+  inputs:
+    - cp_name:
```
mihaitusa: because the "overridable" modifier is true by default the expression should be:  - cp_name: "base"

content/io/cloudslang/createCP/create_package.sl
```diff
@@ -0,0 +1,124 @@
...
+    - cp_location:
+        default: "c:/content_packs"
+        overridable: true
+    - cp_folder: ${cp_location + "/" + cp_name + "-cp-" + cp_version}
+  workflow:
+    - create_Lib_folder:
+        do:
+          create_folder_tree:
+            - folder_name: ${cp_folder + "/Lib"}
+        publish:
```
mihaitusa: I think is better to let go the "publish". Instead maybe is better to put something like: 
navigate:
  SUCCESS: <name_of_the_next_execution_step>
  FAILURE: <SPECIFIC_FAILURE_LABEL_HERE>

mihaitusa:
@talsch  I think it would be very helpful to have also the "deploy_content_pack" operation in "operations_orchestration/content_packs"

mihaitusa:
:+1:

content/io/cloudslang/operations_orchestration/createCP/create_package.sl
```diff
@@ -0,0 +1,126 @@
...
+            SUCCESS: populate_Lib_folder
+            FAILURE: CREATE_LIB_FOLDER_FAILURE
+    - populate_Lib_folder:
+        do:
+          files.write_to_file:
+            - file_path: ${cp_folder + "/Lib/placeHolder"}
+            - text: " "
+        navigate:
+            SUCCESS: create_system_Properties_folder
+            FAILURE: PUPULATE_LIB_FOLDER_FAILURE
```
mihaitusa: (typo) change into: POPULATE_LIB_FOLDER_FAILURE

content/io/cloudslang/operations_orchestration/createCP/create_package.sl
```diff
@@ -0,0 +1,126 @@
...
+        do:
+           files.delete:
+             - source: ${cp_folder}
+        navigate:
+             SUCCESS: SUCCESS
+             FAILURE: CLEAN_FOLDER_FAILURE
+  results:
+    - SUCCESS
+    - CREATE_LIB_FOLDER_FAILURE
+    - PUPULATE_LIB_FOLDER_FAILURE
```
mihaitusa: (typo) change into: POPULATE_LIB_FOLDER_FAILURE

mihaitusa:
@talsch Just 2 typos (please correct them)... after we tested on Linux environment we merge it. :+1:

PR #480, opened 2016-01-05
Chef folder - reviewed comments

SamMarkowitz:


SamMarkowitz:
Still have the following questions:

+ Several flows
  + node_privkey - Is this a path?
  + knife_privkey - What does "(local file that resides where flow is executing)" mean? Same machine? Same folder? Is this a path?
  + knife_host - Is it hostname or IP (like the node_host input)?
+ bootstrap_node.sl
  + node_name (output) - Not documented

Bonczidai:
:ok_hand: 

regarding your questions:

- node_privkey, knife_privkey are paths to key files e.g. /root/keys/yourkey. Consider scenario where you have three machines: local (where you run the CLI), knife server (it has knife installed), node (machine that needs to be managed by knife) and `bootstrap_node` flow. The flow needs a key to access knife server's machine -> `knife_privkey` (this key needs to be on the same machine as the CLI), knife installation needs a key to access the node's machine -> `node_privkey` (this key needs to be on the same machine as knife)

- knife_host ~ node_host - were tested with IP

- node_name is "helper" output, takes its value from node_name input, which is not modified within the flow

PR #482, opened 2016-01-05
DigitalOcean folder - reviewed comments

SamMarkowitz:
Still have the following questions:

+ create_droplet.sl
  + user_data - How are they escaped? Backslash? Enclosing single quotes? Maybe an example would be useful here.
+ wait_for_status_to_change.sl
  + status - Add examples or list of valid values

### DIFF
--- a/content/io/cloudslang/cloud_provider/digital_ocean/v2/droplets/create_droplet.sl
+++ b/content/io/cloudslang/cloud_provider/digital_ocean/v2/droplets/create_droplet.sl
@@ -6,7 +6,7 @@
 #   http://www.apache.org/licenses/LICENSE-2.0
 #
 ########################################################################################################
-# This flow creates a DigitalOcean droplet.
+# Creates a DigitalOcean droplet.
 #
 # Inputs:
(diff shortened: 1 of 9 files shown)


Bonczidai:
:heavy_check_mark: 

regarding your questions:

- user_data: string that contains user data, or with other words, [cloud-config](https://github.com/CloudSlang/cloud-slang-content/blob/aa2a4db07679aaa30e6191fdc047e21e5ff20e59/ci-env/circleci/cloud-config.yaml). this string will be directly inserted into the request body (JSON):
```yaml
{
  "name": "example.com",
  ...
  "user_data": <HERE>,
  "private_networking": null
}
```
so in order to have a valid JSON we need to escape double quotes with a backslash: `"` -> `\"`

- the flow waits until the droplet's status is equal to `status` input - e.g. `new`, the flow waits until the droplet is initializing, it finishes once the droplet's status changes to `active` (or some other status)

SamMarkowitz:
Is there a list of values that are valid for the status inputs? Is this list (from DO docs) correct ["new", "active", "off", "archive"]?

Bonczidai:
yes, they can be considered valid values

PR #487, opened 2016-01-18
Consul folder- reviewed comments

SamMarkowitz:
One question:
  Where there is a default: do we need to keep the `required: false`?


Bonczidai:
IMHO it really depends on the logic of the flow, does the input accept null? (will the flow work if input has null value after binding)

```yaml
- input1:
    default: 'abc'
```
--> `required: true`, if you want `reuqired: false` you have to put it there

@mihaitusa @gituser173 we need to pay attention to this aspect when writing content

Bonczidai:
@SamMarkowitz can this PR be merged? WDYT?

SamMarkowitz:
Yes.

PR #489, opened 2016-01-19
cAdvisor folder - doc review

SamMarkowitz:
Questions:

+ parse_container: has camelCase outputs. Should we change these to snake_case?
+ parse_machine: same as above
+ report_container_metrics: don't understand the description for memory_usage. Could use a rewrite.
+ general: do we need `required: false` on inputs that have a `default`?


gituser173:
@SamMarkowitz let me know if you like the description rewrite for the memory_usage

SamMarkowitz:
Yeah. Looks good.

PR #490, opened 2016-01-19
Docker containers folder - doc review

SamMarkowitz:
Questions:

+ params input: in both `delete_container` and `stop_container` the `params` input is not documented. Is it supposed to be `overridable: false`? If not, please provide a description.
+ standard_err output: in both `delete_container` and `run_container` have and output named `standard_output` whereas most other flows have an output named `error_message`. Should we switch `standard_err` to `error_message`.
+ command input: in both `get_all_containers` and `get_container_ip` the `command` input is not documented. Is it supposed to be `overridable: false`? If not, please provide a description.
+ container_params_cmd: in `start_container` same issue as above

### DIFF
--- a/content/io/cloudslang/docker/containers/clear_container.sl
+++ b/content/io/cloudslang/docker/containers/clear_container.sl
@@ -19,9 +19,6 @@
 #   - port - optional - SSH port
 # Outputs:
 #   - error_message - error message of the operation that failed
-# Results:
-#   - SUCCESS - successful
-#   - FAILURE - otherwise
 ####################################################
 
 namespace: io.cloudslang.docker.containers
--- a/content/io/cloudslang/docker/containers/clear_container.sl
+++ b/content/io/cloudslang/docker/containers/clear_container.sl
@@ -10,12 +10,13 @@
 #
 # Inputs:
 #   - container_id - ID of the container to be deleted
-#   - docker_options - optional - options for the docker environment
+#   - docker_options - optional - options for the Docker environment
 #                               - from the construct: docker [OPTIONS] COMMAND [arg...]
 #   - docker_host - Docker machine host
 #   - docker_username - Docker machine username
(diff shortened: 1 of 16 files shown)


content/io/cloudslang/docker/containers/clear_container.sl
```diff
@@ -10,12 +10,13 @@
...
+#   - docker_options - optional - options for the Docker environment
 #                               - from the construct: docker [OPTIONS] COMMAND [arg...]
 #   - docker_host - Docker machine host
 #   - docker_username - Docker machine username
 #   - docker_password - optional - Docker machine password
 #   - private_key_file - optional - path to private key file
+#   - port - optional - SSH port
 # Outputs:
 #   - error_message - error message of the operation that failed
 # Results:
```
mihaitusa: @gituser173  IMHO the Results section from description can be omitted because are aligned with default behavior. Furthermore the results are not present in the flow.

content/io/cloudslang/docker/containers/delete_container.sl
```diff
@@ -106,8 +106,8 @@ flow:
           FAIL_VALIDATE_SSH: FAILURE
   outputs:
     - result
-    - standard_err
+    - error_message: ${standard_err}
 
   results:
```
mihaitusa: @gituser173  IMHO the Results section from description && flow can be omitted because are aligned with default behavior.

content/io/cloudslang/docker/containers/examples/demo_clear_containers_wrapper.sl
```diff
@@ -67,4 +68,4 @@ flow:
     - error_message
   results:
     - SUCCESS
```
mihaitusa: @gituser173  Same comment as above.

mihaitusa:
@gituser173  It looks good :+1:  Just check to see why build fails.

PR #496, opened 2016-01-22
Adding support sample for Gentoo / CoreOS systems

StefanCristian:


StefanCristian:
@oritstone just a second, I'm pushing another commit, forgot to replace some words :)

PR #497, opened 2016-01-22
Adding support sample for Gentoo / CoreOS systems, part 3

StefanCristian:


StefanCristian:
@mihaitusa all yours :P

PR #506, opened 2016-02-02
OO folder - doc review

SamMarkowitz:
Questions:

+ create_package.sl
  + Should some of the input defaults be removed as they are really just examples?

### DIFF
--- a/content/io/cloudslang/operations_orchestration/createCP/copy_config_items.sl
+++ b/content/io/cloudslang/operations_orchestration/createCP/copy_config_items.sl
@@ -6,16 +6,11 @@
 #   http://www.apache.org/licenses/LICENSE-2.0
 #
 ####################################################
-# Copies yaml configuration files to the system properties folder.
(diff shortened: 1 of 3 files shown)


mihaitusa:
@SamMarkowitz  Actually the "cp_name" && "cp_version" inputs are not optional. The values are needed to successfully deploy it on Central. The name could be anything and the version has to be major.minor.minor (eg. 1.2.3) to have the correct structure that is needed.

SamMarkowitz:
@mihaitusa - Yes, I understand that they are not optional. What I mean is that the default value is not very useful because it is something that you will need to change 90% of the time. So maybe there shouldn't be a default value at all and the field should be plain old required.

mihaitusa:
@SamMarkowitz  I agree with you!

SamMarkowitz:
@mihaitusa - I'll take them out and send it back to you for one final review.

SamMarkowitz:
@mihaitusa - Build is failing, but it seems unrelated.

PR #514, opened 2016-02-10
Fix sp description

Bonczidai:
SP files dopes not have specific description section - should have been skipped during update process

Signed-off-by: Bonczidai Levente levente.bonczidai@hpe.com


gituser173:
:ok_hand:

PR #558, opened 2016-03-29
Strings - add append and length

### LINKED ISSUE #557: Strings - append and length

dashapaster:
fixed #557 Strings - add append and length


mihaitusa:
@dashapaster  Looks good! :+1:

PR #603, opened 2016-04-27
Added YAML operation get_time.sl

scurelea:
Signed-off-by: Curelea Stefan <stefan.curelea@enea.com>


mihaitusa:
@scurelea  Merge the master into your branch and recommit again. The test that fails doesn't have to do anything with your commit but is already fix on master.

content/io/cloudslang/base/datetime/get_time.sl
```diff
@@ -0,0 +1,14 @@
+namespace: io.cloudslang.base.datetime
```
mihaitusa: @scurelea Add the description section like in the other content files.

PR #621, opened 2016-05-05
Reorganize base content

### LINKED ISSUE #616: location of some base content is missleading
1.	HTTP client operations were searched under base/http folder but they are actually under base/network/rest folder. What is the motivation of having them under base/network/rest folder? https://github.com/CloudSlang/cloud-slang-content/tree/master/content/io/cloudslang/base/network/rest 

2.	SSH should be under base/ssh folder. Is there a specific reason why they are under remote_command_execution folder?

3.	‘cmd’ why do we have a dedicated folder called ‘base/cmd’ for one operation that starts a process?

4.	check_ssh_unstable_session is located under base/utils should be under base/ssh folder

SamMarkowitz:
fixes #616 

Leaving cmd folder as is for now. Not sure where else to put the `run_command` operation.

### DIFF
--- a/content/io/cloudslang/base/http/http_client_action.sl
+++ b/content/io/cloudslang/base/http/http_client_action.sl
@@ -96,7 +96,7 @@
 #!!#
 ################################################
 
-namespace: io.cloudslang.base.network.rest
+namespace: io.cloudslang.base.http
 
 operation:
   name: http_client_action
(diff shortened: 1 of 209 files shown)


mihaitusa:
@SamMarkowitz  Everything seems ok! :+1:

lrevnic:
@SamMarkowitz score-actions location of the operations is still the old one. It seems the work was done in cloudslang-content project only but not in score-actions project. They should be aligned.

meshipeer:
@lrevnic the actions are not CS - but score... they can also be used in other languages supported by score engine (like AFL..)

PR #622, opened 2016-05-06
Added 3 tests in trim.input.yaml & some small changes in search_and_replace.inputs.yaml file

andreeapuscas:
the following tests were added in trim.input.yaml: testTrimNullValueSuccess, testMissingInputs and testTrimWhitespacesInsideStringSuccess

Signed-off-by: Andreea Puscas  andreea.puscas@hpe.com

### DIFF
--- a/test/io/cloudslang/base/strings/trim.input.yaml
+++ b/test/io/cloudslang/base/strings/trim.input.yaml
@@ -36,18 +36,16 @@ testTrimBeforeSuccess:
 testTrimNullValueSuccess:
   inputs:
     - string: ''
-  description: Tests that trim.sl finishes with success when input is a null value
+  description: Tests that trim.sl finishes with success when input is empty string
   testFlowPath: io.cloudslang.base.strings.trim
   outputs:
     - result: ''
   result: SUCCESS
 
 testMissingInput:
-  inputs:
+  inputs: []
   description: Tests that trim.sl operation throws an exception when the required input is missing
   testFlowPath: io.cloudslang.base.strings.trim
-  outputs:
-    - result: test
   throwsException: true
 
 testTrimWhitespacesInsideStringSuccess:
--- a/test/io/cloudslang/base/strings/trim.input.yaml
+++ b/test/io/cloudslang/base/strings/trim.input.yaml
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
   outputs:
     - result: good
   result: SUCCESS
+
+testTrimNullValueSuccess:
+  inputs:
+    - string: ''
+  description: Tests that trim.sl finishes with success when input is a null value
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
+    - result: ''
+  result: SUCCESS
+
+testMissingInput:
+  inputs:
+  description: Tests that trim.sl operation throws an exception when the required input is missing
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
+    - result: test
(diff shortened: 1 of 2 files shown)


test/io/cloudslang/base/strings/trim.input.yaml
```diff
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
...
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
+    - result: ''
+  result: SUCCESS
+
+testMissingInput:
+  inputs:
+  description: Tests that trim.sl operation throws an exception when the required input is missing
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
```
Bonczidai: do you need `outputs` here if test is supposed to throw an exception?

test/io/cloudslang/base/strings/trim.input.yaml
```diff
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
...
+  inputs:
+    - string: ''
+  description: Tests that trim.sl finishes with success when input is a null value
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
+    - result: ''
+  result: SUCCESS
+
+testMissingInput:
+  inputs:
```
Bonczidai: with this syntax inputs have `null` value, right? (since it's Yaml) can we skip inputs tag or use `inputs: []` syntax?

test/io/cloudslang/base/strings/trim.input.yaml
```diff
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
   outputs:
     - result: good
   result: SUCCESS
+
+testTrimNullValueSuccess:
+  inputs:
+    - string: ''
+  description: Tests that trim.sl finishes with success when input is a null value
```
Bonczidai: input is empty string, not null in the example

test/io/cloudslang/base/strings/trim.input.yaml
```diff
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
   outputs:
     - result: good
   result: SUCCESS
+
+testTrimNullValueSuccess:
+  inputs:
+    - string: ''
```
Bonczidai: does the `trim` op have `string` as the input name? that's error prone -- just like these values: https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt

test/io/cloudslang/base/strings/trim.input.yaml
```diff
@@ -32,3 +32,29 @@ testTrimBeforeSuccess:
...
+  description: Tests that trim.sl finishes with success when input is a null value
+  testFlowPath: io.cloudslang.base.strings.trim
+  outputs:
+    - result: ''
```
Bonczidai: same opinion for `result` - seems error prone to me, especially because in CloudSlang result means flow / op result (resolution)

Bonczidai:
build fails -- from the logs:

```
01:21:27 [ERROR] - Error occurred while running test: testSearchAndReplaceUsingNumbers - Tests that search_and_replace.sl operation finishes with SUCCESS when inputs contain digits
		Error occurred during navigation execution  , Error executing python script: Traceback (most recent call last):
	  File "<string>", line 2, in <module>
	TypeError: coercing to Unicode: need string or buffer, int found
```

Bonczidai:
:ok_hand:

PR #619, opened 2016-05-05
added CS operations and flows for date/time functionality

cdemergian:
fixes #603

PR #629, opened 2016-05-12
Date && Time operations

mihaitusa:
fixes #619


mihaitusa:
@cdemergian  After we perform a release to MC we can rebuild this without problems. So, until that just check to see the modifications.

mihaitusa:
Build issue related to "score-" concatenation in naming....

PR #637, opened 2016-05-17
Version updates - v0.9.58

Bonczidai:
fixes #636 

Signed-off-by: Bonczidai Levente levente.bonczidai@hpe.com

PR #645, opened 2016-05-20
Input output different names

mihaitusa:
align with latest DSL changes


Bonczidai:
migrated to #637 - thx

PR #650, opened 2016-05-26
Sleep.sl - enhancements

tethryus:
Added FAILURE step
Treats exceptions
Changed behavior
Added FAILURE tests

Signed-off-by: Sorin Moldovan <sorin.moldovan@hpe.com>


SamMarkowitz:
Looks good.

PR #652, opened 2016-05-26
added sensitive property

SamMarkowitz:
fixes CloudSlang/cloud-slang#467


SamMarkowitz:
@mihaitusa, if you have a minute, please take a look to see that I have the right idea.

SamMarkowitz:
Build errors as in #653. Waiting on their resolution.

content/io/cloudslang/base/http/http_client_action.sl
```diff
@@ -115,43 +115,55 @@ operation:
         private: true
     - username:
         required: false
+        sensitive: true
```
Bonczidai: together with @mihaitusa we think only password fields like `password`, `proxy_password`, `trustPassword` should be sensitive -- is there a specific reason making other values sensitive? (is this a convention we are going to use?)

SamMarkowitz: Got it. Will remove the others.

SamMarkowitz:
@mihaitusa I have one build error still. I'm not sure what's causing it. Can you take a look?

SamMarkowitz:
Strange, I was sure I did that already, but it turns out I didn't. Let's hope it works.

PR #674, opened 2016-06-16
Added tests in add_entry_to_object.inputs.yaml file.

andreeapuscas:
Signed-off-by: Andreea Puscas <andreea.puscas@hpe.com>


mihaitusa:
@andreeapuscas Please replace the existing tests with wrapper flows that verifies certain blocks within a JSON. The order within JSON object is not preserved. We cannot merge this PR.

PR #710, opened 2016-07-14
changed some operations to decisions

### LINKED ISSUE #706: change relevant content to be desicion

SamMarkowitz:
fixes #706


Bonczidai:
:+1: 

IMHO decisionis like decision in a flow chart - base don this, only the following are considered decisions from the PR: equals, compare_lists,  compare_numbers, is_true (the other like math operations are not)

you could also add:
- `content/io/cloudslang/base/lists/contains.sl` you should use comma as default delimiter (the op is not correct right now) that way you'll only ned this expr: `element in list.split(delimiter)`
- `content/io/cloudslang/base/strings/string_equals.sl`

SamMarkowitz:
`string_equals` was a bit awkward to make into a decision. Maybe you have a better idea of how to do it?

Also, `lists/contains` seems to have switched to a `java_action`. Not sure why, but I left it as such.

(Lastly, I had to restart the build three times to get it green.)

Bonczidai:
:+1: 

17qmv5

PR #714, opened 2016-07-18
Fix tests after builder updated

mihaitusa:
at version 0.9.63


mihaitusa:
This PR was made just for investigation of a certain issue... so it can be safely closed.

PR #722, opened 2016-07-20
Update http

tethryus:
Updated GAV to reflecr name change from score to cs


tethryus:
@mihaitusa Please review

PR #724, opened 2016-07-20
Update lists

tethryus:
GAV update
contains.sl changed from python to java
Fixed tests to reflect behavior change


content/io/cloudslang/base/lists/contains.sl
```diff
@@ -7,12 +7,21 @@
 #
 ####################################################
 #!!
-#! @description: Check element in the list.
-#! @input list: list in which we need to check the element  - Example: [123, 'xyz']
-#! @input delimiter: list delimiter
-#!                   default: ''
-#! @output element: element which we want to check
-#! @output result: true - if list contains, false - if not
+#! @description: This operation checks to see if a list contains every element in another list
```
mihaitusa: This operation checks to see if a list contains every element **from/of** (... pick one) another list

content/io/cloudslang/base/lists/contains.sl
```diff
@@ -7,12 +7,21 @@
...
+#! @input container: the containing list  - Example: Luke,Vader,Kenobi
+#! @input sublist: the contained list - Example: Kenobi
+#! @input delimiter: a delimiter separating elements in the two lists
+#!                   default: ','
+#! @input ignore_case: If set to 'true' then the compare is not case sensitive.
+#!                     default: 'true'
+#! @output response: 'true' if found, 'false' if not found
+#! @output return_result: empty if sublist found in container
```
mihaitusa: @tethryus  That's the way we want it? Nothing on return_result on success? Not even a meaningful message?... If we want it that way then leave it as it is....

content/io/cloudslang/base/lists/contains.sl
tethryus: from it is, thank you,

content/io/cloudslang/base/lists/contains.sl
tethryus: That is the expected behavior, if alll elements from the sublist were found in the container, then return_result will be empty and the response will be 'true'. In case an element from the sublist is not found in the container, then return_result will be populated with those values and response false.

PR #726, opened 2016-07-21
Update SSH

tethryus:
GAV update to reflect the change from score to cs.


mihaitusa:
@tethryus Please see why the build fails.

tethryus:
Issue fixed, all should be ok now.

Issue #589, opened 2016-04-18
Improve base/cmd/run_command


Bonczidai:
Run_command from base/cmd folder can be improved so that STDOUT and STDERR will be outputted too

from @andreivasiletruta

Issue #150, opened 2015-05-13
run_cmd add cwd param


meirwah:
add a file directory input param to run the cmd from... 
to operation run_command

PR #742, opened 2016-08-02
Run command enhancement

tethryus:
Closes: #589 
Closes: #150 
The command has:
 return_result for the command output
 error_message if there is an error with the command
 return_code with the significant return codes in case of SUCCESS or FAILURE
Added CWD 
Added STDOUT
Added STDERR


mihaitusa:
@tethryus  testRemoteSecureCopySucessLocalToRemote still fails... please check it!

PR #744, opened 2016-08-02
create_folder - enhancement

tethryus:
operation uses makedirs instead of mkdir
operation supports environment variables
examples for each case in description


mihaitusa:
@tethryus  Check out this test: testCreateFolderFailureIllegal

PR #753, opened 2016-08-05
added test on strings for input bool_value

razvan-lupu:
The operation doesn't work on central because in central the input is given as String. I added also a comparison for strings in order to work both on central and CloudSlang-cli.

### DIFF
--- a/content/io/cloudslang/base/utils/is_true.sl
+++ b/content/io/cloudslang/base/utils/is_true.sl
@@ -21,5 +21,5 @@ decision:
   inputs:
     - bool_value
   results:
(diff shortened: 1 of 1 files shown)


content/io/cloudslang/base/utils/is_true.sl
```diff
@@ -21,5 +21,5 @@ decision:
   inputs:
     - bool_value
   results:
-    - SUCCESS: ${ bool_value == True }
+    - SUCCESS: ${ bool_value == True or bool_value == "True" }
```
mihaitusa: @razvan-lupu You should consider this: bool_value in [True, true, 'True', 'true'] to cover up all the possibilities. WDYT?

PR #754, opened 2016-08-08
Regex replace enhancement

Tirla-Alin:


mihaitusa:
@Tirla-Alin  You should update the description section too... according with your change

PR #756, opened 2016-08-09
Fix string equals

razvan-lupu:
Fixes ignore case problem in string_equals.sl by allowing multiple values for boolean "true".

Signed-off-by: Razvan Lupu razvan.lupu@hpe.com


mihaitusa:
@razvan-lupu  In order to better communicate between us please use the appropriate label... which is "in progress" as long as you work on the PR. When build passes and work is done on it then remove the "in progress" label, put "ready for merge" label and assign one of us for review/merge.

PR #764, opened 2016-08-17
added json_xml_converters

victorursan:
Signed off: Victor Ursan <victor.ursan@hpe.com>


mihaitusa:
@victorursan -> Test case failed: testSelectBadXpathInvalidXmlDocument - Tests that this operation fails when xml_document input is invalid. Flow xpath_query **did not throw an exception as expected** It seems that behavior changed.

PR #774, opened 2016-08-22
Content changes

tethryus:
Renamed files to file_system
Moved certain operations under http
Renamed os to operating_systems
Renamed scripts to python
Modified test paths


chronical:
few comments: 
- operating_systems should be changed to os as it's a well known acronym
- rrt is confusing for a new user, we should get rid of this subfolder and move the remove file system operations in the filesystem folder

tethryus:
PR merged in WS_1_0

PR #785, opened 2016-08-31
Lists - sort list fix

tethryus:
Fixed bug where "reverse" input in sort_list.sl was supposed to be a string (due to recent changes) but was Boolean.
Updated GAV version in all lists operations that use Java.

Signed-off by Sorin Moldovan <sorin.moldovan@hpe.com>


Bonczidai:
:+1:

PR #858, opened 2016-10-21
Refactored sls

Tirla-Alin:
endpoint required false, default value
identity required true, sensitive false
credential required true, sensitive true
delimiter required false, default value


tethryus:
@Tirla-Alin Looks ok 👍 ready for merge.

PR #887, opened 2016-11-21
Amazon

alinaciobanca:
- remove region input
- map proxy_username and proxy_password


Bonczidai:
:+1: 

please make sure to fulfill the following guidelines in the future:
- commits should be signed
- PR description should be signed
(you can take a look on other PRs for a reference)

PR #893, opened 2016-11-29
removed defaults from mutually exclusive inputs

victorursan:
p.s. does @note work ?

Signed-off by: Victor Ursan <victor.ursan@hpe.com>


tethryus:
@victorursan At the moment, @note does not work, but it will be implemented in the future. Leave it as it is.

PR #896, opened 2016-12-15
Refactored AWS start/stop/reboot for CSA compatibility

Madalin-HPE:
Refactored Amazon AWS start/stop/reboot e2e flows for CSA compatibility issue with results longer than 4000 characters.


tethryus:
@Madalin-HPE 👍  Can be merged.

PR #920, opened 2017-01-20
Azure:

alinaciobanca:
- changed the range for random_number_generator - to always generate a 5 digits number
- reset the value of vm_name output in case of flow failure
- removed useless lines

Signed-off-by: Alina Ciobanca <alina.ciobanca@hpe.com>


tethryus:
👍

PR #922, opened 2017-01-25
fixing issue #966 

pe-pan:
Closes: #923 

Two major changes:
- the bool_value argument of utils.is_true needs to be converted to string before evaluation
- at verify_group_exist flow, the command_return_code is not 0 when the group does not exist


Bonczidai:
👌

PR #935, opened 2017-02-15
Update .travis.yml for slack notification

andreivasiletruta:
For the purpose of using slack based communication I want to add the travis related keys so that will get notified for test based results.


andreivasiletruta:
closing until further investigation is done on the forked repository

Issue #933, opened 2017-02-15
Yaml "default: None" gives unicode 'None' to python script


FlorianJDF:
Hello,

After trying to understand why the parameter cwd in the flow run_command.sl was always ignored, I found this problem : 
If a parameter has a default value None in the Yaml, the value of this parameter is the unicode None. So the condition "is None" is always false.

Here is the code and the result:
---------------------------------------------------------------------------------------------
```
operation:
  name: run_command
  inputs:
    - command
    - cwd:
        required: false
        default: None

  python_action:
    script: |
      import os
      import subprocess
      print "value cwd: " + cwd + " type:" + str(type(cwd))
      print (cwd is None)
[...]
```
---------------------------------------------------------------------------------------------
$ value cwd: None type:<type 'unicode'>
$ False
----------------------------------------------------------------------------------------------

Regards,

Bonczidai:
@FlorianJDF this is true

`- default: val` -- if `val` has the format `${expr}` then `expr` will be sent to the python interpreter, otherwise it is considered as a string value; hence `default: None` ~ `default: 'None'` ~ `default: "None"` (this is from Yaml) 

you can achieve the desired behavior with `default: null` or eventually `default: ${None}`

if you have a **fix** for this op please contribute it back 💪 

@tethryus WDYT the default value should be in this case for `cwd`?

tethryus:
To maintain the same behavior, cwd default value: should be - ' ' - (empty string) and change the behavior of the operation, in case cwd by default is an empty string.

Bonczidai:
adding @dandenisa

FlorianJDF:
@Bonczidai Thank you for your answer, i'll send you a contribution

PR #936, opened 2017-02-16
Modification of the cmd flow, in order to give None to the python script and to overwrite the var cwd only if not provided by the user

FlorianJDF:
Closes: #933 

The default value is now "null" in order to get None in the python script.
The condition to overwrite cwd is changed in order to do a os.getcwd() if cwd is null.


tethryus:
@FlorianJDF Looks good. 👍  Can be merged. Thanks for the contribution.
@Bonczidai WDYT?

PR #947, opened 2017-03-02
New Google Cloud operation: Instance Set Metadata

Tirla-Alin:
Created SL and test SL for the operation

o	Signed-off by Tirla Alin alin.tirla@hpe.com


tethryus:
Continued in PR #948

PR #1011, opened 2018-01-19
Local ping

pintea-eugen:
Added .sl file for the cs-utilities io..cloudslang.content.actions.LocalPing Action
Added tests (one positive and negative to test the implementation).
The tests are running at each build, which it also downloads the dependencies needed to run the operation.

Signed-off-by: Eugen Pintea, <pintea.,eugen@hpe.com>


tethryus:
@pintea-eugen If the build is green, the PR can be merged. The tests that come with the operation will run.

PR #1029, opened 2018-02-07
adding delete_secret.sl

andreivasiletruta:
Signed-off-by: Andrei Vasile Truta <andrei-vasile.truta@hpe.com>


tethryus:
@robert-sandor WDYT?

PR #1031, opened 2018-02-07
Feature/slack

andreivasiletruta:
adding post_message_raw so that one can use cloudslang oo to post messages (with attachments) on a slack channel/private message/instant message.


andreivasiletruta:
@tethryus thanks for the updates. should we drag @robert-sandor for small review?

tethryus:
yeah, @robert-sandor WDYT?

PR #1057, opened 2018-06-14
update cs-amazon version

lrevnic:
add example workflow
update cs-action version


lrevnic:
Please wait with the review until I add description :)

PR #1063, opened 2018-08-01
Changed validate_json.sl to wotk with Java action

danielmanciu:
Changed validate_json.sl to work with Java action
Added more tests


content/io/cloudslang/base/json/validate_json.sl
```diff
@@ -1,28 +1,26 @@
-#   (c) Copyright 2017 EntIT Software LLC, a Micro Focus company, L.P.
```
lrevnic: Should be 2018 Micro Focus

PR #1114, opened 2018-10-31
handle hardcoded delimiters if exist in the key/values of the uft parameters

danielcrisan:
- handle comma escape in the parameters keys or values
- add parameter for name_value delimiter


lrevnic:
@danielcrisan in general we use , as default value for delimiter; what is the reason for not following the same practice here?

danielcrisan:
In this use case we talk about two delimiters in the same input because parameters inputs is a list of key:value. So about the delimiter between the key and value, user should be able to change it because it can happen that the value contains ":" character. Is the same with the , delimiter for the list, user should be able to escape the , in the param value. Because , is already used for the list iteration.

Issue #1118, opened 2018-12-14
Operation io.cloudslang.base.lists.add_element.sl does not allow empty input parameters (empty list, empty element)


pe-pan:
I'm iterating through a json array and filtering it to a list. The list is empty at the beginning but unfortunately, the add_element operation does not allow empty list as an input parameter. IMHO, both inputs - list and element should be allowed to be empty (optional).

PR #1119, opened 2018-12-14
Fix add element

pe-pan:
Fixing issue #1118


ionutmarisca:
The new Python implementation looks really good to me as it's really simple and easy to understand. 

However, I would recommend creating a new flow instead of modifying the existing one so we can avoid any backwards compatibility issues. 

I think adding a new `Python` folder under `cs-content/content/io/cloudslang/base/lists/` and creating the new .sl file there should be fine. 

In the future, after proper testing of the new `Python` implementation we can migrate all the other flows to it.

PR #1180, opened 2019-12-24
Terraform applydetails

manjoosh12:
Created an SL file for applies and tested end to end.


content/io/cloudslang/hashicorp/terraform/runs/get_apply_details.sl
```diff
@@ -0,0 +1,206 @@
...
+
+operation: 
+  name: get_apply_details
+  
+  inputs: 
+    - auth_token:    
+        sensitive: true
+    - authToken: 
+        default: ${get('auth_token', '')}  
+        required: false
```
KalyankarRajesh: Remove required: false for both authToken and applyId

manjoosh12: updated

PR #1185, opened 2020-01-09
Feature/terraform variables

manjoosh12:
Created a SL file for Cancel Run and tested it.


manjoosh12:
already handled in some other PR

PR #1204, opened 2020-05-07
Changed the gav version in all Terraform Operations

MeghanaManjunath98:
Changed the gav version in all Terraform Operations


content/io/cloudslang/hashicorp/terraform/workspaces/variables/create_workspace_variable.sl
```diff
@@ -0,0 +1,243 @@
+########################################################################################################################
```
patmante, line 1: add copyright message

MeghanaManjunath98: Added the copyright message

PR #1245, opened 2020-07-01
Added operations for detach vnic

KalyankarRajesh:
Added operations for detach vnic and workflow for detach vnic


KalyankarRajesh:
Closing PR

PR #1303, opened 2020-10-08
added the get_epoch_time operation

florinAndrei1991:
added the get_epoch_time operation to io.cloudslang.base.datetime


florinAndrei1991:
Seems to me that regardless of the test suite the tests are added on, the PR checks will not run the tests.
Could you please advise?

ericandreica (approved):

PR #1329, opened 2021-03-10
Office365 add attachment

florinAndrei1991:


mateiioana925 (approved):

content/io/cloudslang/microsoft/office365/send_email.sl
```diff
@@ -45,6 +45,8 @@
 #!                 Optional
 #! @input body: The body of the message. Updatable only if 'isDraft' = true.
 #!              Optional
+#! @input file_path: The absolute path to the file that will be attached. This input is mutually exclusive with the "content_name" and "content_bytes" inputs.
```
lrevnic, line 48: It will be good to mention/add an example of how the path looks like if executed from inside containers.
Also, if there are any constraints/limits on the length of the paths or size of files etc.

PR #1330, opened 2021-03-12
Edited the upload_object.sl flow

IuliaMoldovan:


content/io/cloudslang/google/storage/objects/upload_object.sl
```diff
@@ -152,7 +134,7 @@ flow:
             - connect_timeout
             - socket_timeout
             - content_type: application/json
-            - headers: "${'Authorization: Bearer ' + access_token}"
+            - headers
```
ericandreica, line 137: it works without token?

IuliaMoldovan: the "headers" input is now exposed in the flow so now the user should set the Authorization type

ericandreica (approved):

PR #1400, opened 2021-08-27
Added public_ip_address output property

patmante:
Created following workflows and added public_ip_address as output property
* start_vm_v3.sl
* stop_and_deallocate_vm_v3.sl

Tested the flows with public and private VMs


umavanth21 (changes requested):
"This workflow stops and deallocates the virtual machine."
Change the description to "This workflow stops the virtual machine and deallocates the Public IP address from the virtual machine."

umavanth21 (approved):
Approved

ameydwivedi (approved):

PR #1418, opened 2021-09-13
Added Octane Integration

AndreiCiurba:
Added operations and flows for the Octane integration (containing operations and flows for authenticating and for creating entities such as defects, releases, workspaces and users).


andinomm (approved):
It is looking good!

ioana-matei (approved):

PR #1426, opened 2021-09-21
Undeploy gcp workflow 

manjoosh12:
1>Created a new undeploy work flow to delete the instances from the gcp console.
2>modified the  output order of insert instance operation is cause of binding output issues at workflow designer.

### DIFF
--- a/content/io/cloudslang/google/compute/compute_engine/instances/insert_instance.sl
+++ b/content/io/cloudslang/google/compute/compute_engine/instances/insert_instance.sl
@@ -13,7 +13,7 @@
 #
 ########################################################################################################################
 #!!
-#! @description: Creates an instance resource in the specified project using the data included as inputs.
+#! @description: Creates an instance in the specified project using the data included as inputs.
 #!
 #! @input access_token: The access token from get_access_token.
 #! @input project_id: Google Cloud project name.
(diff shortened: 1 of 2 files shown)


sushilMF (changes requested):
for insert_instance.sl file change the copyright

manjoosh12:
> for insert_instance.sl file change the copyright

@sushilMF  updated .

umavanth21 (commented):
@manjoosh12 , what do u mean by " if async is false"

#! @output instance_id: The id of the instance if async is false.
#! @output internal_ips: The internal IPs of the instance if async is false.
#! @output external_ips: The external IPs of the instance if async is false.

manjoosh12:
> @manjoosh12 , what do u mean by " if async is false"
> 
> #! @output instance_id: The id of the instance if async is false.
> #! @output internal_ips: The internal IPs of the instance if async is false.
> #! @output external_ips: The external IPs of the instance if async is false.

@umavanth  there is a input property called asnyc when it set it to "false" during deploy then we get instance_id, internal_ips, external_ips values . it it set to true then we dont get these output values.

umavanth21:
The description is confusing. It must convey like below (pls check with nandini)
instance_id: The id of the instance. When async is set to false, this value is updated.

sushilMF (changes requested):
Do we need timeout at workflow level in undeploy?
image

manjoosh12:
> Do we need timeout at workflow level in undeploy?
> image

@sushilMF "yes we need it"

sushilMF (approved):

umavanth21 (approved):
Thanks for updating the description

PR #1607, opened 2022-08-11
Created a operation for Delete instance and Restart instance

manjoosh12:
Created a operation for Delete instance and Restart operation


umavanth21 (changes requested):
Please change the description like below:
zone: The zone in which the instance resides.

manjoosh12:
> Please change the description like below: zone: The zone in which the instance resides.

updated the description @umavanth

umavanth21 (approved):

KalyankarRajesh (approved):

PR #1611, opened 2022-09-05
Added a operation for to attach disk ,detach disk,insert disk,delete disk,insert instance,get subnetworks

manjoosh12:
Added a operation for to attach disk ,detach disk,insert disk,delete disk,insert instance,get subnetworks


umavanth21 (approved):
Thanks Manjoosh for checking in. If there is any change required, we will take care.

KalyankarRajesh (approved):

PR #1627, opened 2022-11-15
Added the create tag operation

ghost:
Created "create tag user operation in microsoft azure"


umavanth21 (changes requested):
- Include MicroFocus 2022(refer other operations)
- Is it authentication token or authorization token?
- namespace is wrong
- why all values are hardcoded like subscription ID, token, proxy?
- default values mentioned as allow_all and true for security properties

PR #1629, opened 2022-11-16
Created operation to create or add tags to a resource in Azure and ad…

ArthyVenk:
…ded a python flow to convert the comma separated values to JSON format


KalyankarRajesh (changes requested):
Update the namespace and property description.

umavanth21 (commented):
- check line no 79 -is it pointing to test?
- auth token - please follow the same definition as used in other flows
- keys and values - can we say tag name and tag value , please revisit this description

KalyankarRajesh (approved):

umavanth21 (approved):

PR #1657, opened 2023-01-06
Added operation a new operation

ArthyVenk:
Added operation to check if label keys and label values are equal in the deployVm flow


umavanth21 (changes requested):
Line no164 points to your local folder

umavanth21 (approved):

sushilMF (approved):

PR #1668, opened 2023-02-16
New AWS Deploy Instance Flow with expose tags feature

Sailakshith6:
New AWS Deploy Instance Flow with expose tags feature


umavanth21 (commented):
Existent network interfaces can be attached or, if desired, one or more new network interfaces can be created and attached to the instance.  - what it means? Are we providing this functionality?

KalyankarRajesh (approved):

umavanth21 (approved):

